PR #6, opened 2018-12-12
Switched to Debian Jessie as a base container

clintonb:
Alpine builds break too frequently. While the image size may be smaller, the inconvenience is not worth the benefit.


r-b-g-b:
Any reason why preference for Jessie over Stretch?

clintonb:
I've never heard of stretch.

r-b-g-b (approved):

PR #5, opened 2018-12-12
Added support for PostGIS

clintonb:
Based on #6 (Docker base image changes)

PR #9, opened 2018-12-13
Added support for PostGIS

clintonb:
This is a redux of #5 and #6. The primary difference is the first commit, which updates the Travis build.

The Travis build was previously failing, with processes exiting with code 137. This exit code usually indicates the process was killed because the system ran out of memory. Since the processes were running in the `app` container, the only other process that would kill the memory is `gunicorn`.

That process is no longer run on Travis. Instead we restart the containers for every step of the build script, each time running only the process necessary to complete that part of the script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-sudo: required
-
 services:
   - docker
 
@@ -13,12 +11,11 @@ cache: pip
 install:
   - bash .travis/install_terraform.sh
   - make docker.build
-  - make local.up
 
 script:
-  - docker exec -t woeip.app make quality
-  - docker exec -t woeip.app make static
-  - docker exec -t woeip.app make test
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "make quality"
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "make static"
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "make test"
   - make validate_terraform
 
 after_success:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6-alpine
+FROM python:3.6-jessie
 
 WORKDIR /app/woeip
 
@@ -6,16 +6,14 @@ ENV DJANGO_SETTINGS_MODULE woeip.settings
 ENV PIPENV_DONT_USE_PYENV 1
 ENV PIPENV_SYSTEM 1
 
-RUN apk add --update \
-    coreutils \
-    gcc \
-    libffi-dev \
-    make \
-    musl-dev \
-    postgresql-dev \
-    python3-dev \
-  && pip install pipenv \
-  && rm -rf /var/cache/apk/*
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y \
+        build-essential \
+        gettext \
+        libffi-dev \
+        libssl-dev \
+    && rm -rf /var/lib/apt/lists/* \
+    && pip install pipenv
 
 COPY Makefile /app/woeip
 COPY Pipfile /app/woeip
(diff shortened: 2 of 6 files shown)


clintonb:
Feel free to "Rebase and merge" when ready.

.travis.yml
```diff
@@ -13,12 +11,11 @@ cache: pip
 install:
   - bash .travis/install_terraform.sh
   - make docker.build
-  - make local.up
 
 script:
-  - docker exec -t woeip.app make quality
-  - docker exec -t woeip.app make static
-  - docker exec -t woeip.app make test
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "make quality"
```
TangoYankee, line 16: Interesting way to solve the problem: create a compose configuration specifically for travis. Noting the difference between production and travis compose will help teach me a little more about both- and exactly how gunicorn does its thing.

Dockerfile
```diff
@@ -1,21 +1,20 @@
...
-    coreutils \
-    gcc \
-    libffi-dev \
-    make \
-    musl-dev \
-    postgresql-dev \
-    python3-dev \
-  && pip install pipenv \
-  && rm -rf /var/cache/apk/*
+RUN apt-get update && \
```
TangoYankee, line 9: The bash commands are definitely welcome sight. I'm more familiar with debian/ubuntu. An unintended consequence of switching from alpine is that its making it easier for me to decode how this all works.

Makefile
```diff
@@ -36,7 +36,7 @@ production-requirements: ## Install requirements for production
 	pipenv install
 
 test: clean ## Run tests and generate coverage report
-	SECRET_KEY=fake DATABASE_URL="sqlite://:memory:" coverage run -m pytest --durations=25 -v
+	coverage run -m pytest --durations=25 -v
```
TangoYankee, line 39: From a quick glance, this appears to mean that the test database will switch from sqlite (which would've needed to be spatialite to support gis capabilities) to postgis

docker-compose.local.yml
```diff
@@ -17,12 +17,12 @@ services:
   app:
     # Instruct Gunicorn to reload code when it is changed locally.
     command: --reload
-    links:
-      - db
     environment:
       - DEBUG=true
       - SECRET_KEY=replace-me
-      - DATABASE_URL=psql://postgres:postgres@woeip.db:5432/woeip?connect_timeout=60
+      - DATABASE_URL=postgis://postgres:postgres@woeip.db:5432/woeip?connect_timeout=60
```
TangoYankee, line 23: Obligatory refractoring, to upgrade from gres to gis

woeip/settings.py
```diff
@@ -26,6 +26,7 @@
     'django.contrib.messages',
     'django.contrib.staticfiles',
     'django.contrib.flatpages',
+    'django.contrib.gis',
```
TangoYankee, line 29: Looks like we're geodjango enabled too

TangoYankee (approved):
Nothing to change. The comments are there mostly to communicate my thought process as I reviewed it.

TangoYankee:
I'd like to give @r-b-g-b  and @YotamHa @ethan-bradley a chance to review it while it's still a pull request. I think pull requests are a good pacing mark, to help everyone learn the code base together. After all, this is a lot for us to learn.

r-b-g-b (approved):
All looks good to me.

Makefile
clintonb: Correct. The tests now use whatever database is configured via `DATABASE_URL`. This is ideal because you avoid potentially having tests that work for one database backend but not another. As the project grows, you may wish to go back to using SQLite for speed. Just be aware of the tradeoff.

clintonb:
I'm merging now. You can continue posting comments and questions to PRs after merging, so keep them coming. I'll answer as time allows.

PR #10, opened 2018-12-16
BUG: local setup used postgres, switch to mdillon/postgis

r-b-g-b:
Minor change, we forgot to make the local configuration use postgis.


docker-compose.local.yml
```diff
@@ -28,5 +28,7 @@ services:
       - 8000:8000
     volumes:
       - .:/app/woeip:cached
+    depends_on:
```
clintonb: Nit: alphabetize these

clintonb (approved):

PR #11, opened 2018-12-16
WIP: Add database to the app

r-b-g-b:
Added the database structure to our app. The structure was outlined in the "WOEIP database structure" doc (find it on ``/docs list`` on the slack channel).


woeip/apps/core/models.py
```diff
@@ -1,6 +1,123 @@
...
+    name = models.CharField(max_length=256)
+    website = models.CharField(max_length=256)
+    email = models.CharField(max_length=256)
+    contact = models.ForeignKey(User, on_delete=models.SET_NULL, blank=True, null=True)
+
+    def __str__(self):
+        return f"{self.name} <{self.email}>"
+
+
+class RoleEnum(models.Model):
```
clintonb: Take a look at Django’s existing permissions system. There are also row-based permissions libraries you can use.

r-b-g-b:
^^ Build failing Q: I get this error locally if I don't ``./manage.py makemigrations core && ./manage.py migrate``. I could invent some process to do this within the cookiecutter framework, but I bet something like that already exists @clintonb ?

woeip/apps/core/models.py
```diff
@@ -1,6 +1,123 @@
...
+        if self.can_delete:
+            crud.append('delete')
+
+        crud = '/'.join(crud)
+
+        return f"{self.name} ({crud})"
+
+
+class Participant(models.Model):
+    """A participant"""
```
clintonb: Useless comment is useless.

woeip/apps/core/models.py
```diff
@@ -1,6 +1,123 @@
...
+
+class Device(models.Model):
+    name = models.CharField(max_length=256)
+    manufacturer = models.CharField(max_length=256)
+    serial_number = models.CharField(max_length=256)
+    model_number = models.CharField(max_length=256)
+    calibration_date = models.DateField()
+    firmware_version = models.CharField(max_length=256)
+
+    def __str__(self):
```
clintonb: You don’t need this for every model. In this case, you’re adding so much information, that the user might as well just look at all of the model fields.

clintonb:
What error do you get? If you don’t have a virtualenv, open a shell into the container and run the commands there. See the Makefile for the shell command.

r-b-g-b:
Some topics for discussion:
 - ``on_delete`` settings: when two models are linked by a foreign key, it needs to be decided what happens when one of the entries is deleted. For example, Participant and Organization are linked, so what should happen when an org is deleted? Should that delete all of that org's participants (``models.CASCADE``)? Or should those participants' orgs field be set to null (``models.SET_NULL``). I made some provisional decisions just to get things running. (For the participant/org case, I decided to use "set null".)
 - testing: we should decide on how to test this. There's this [post](https://stackoverflow.com/questions/43708419/how-to-prepopulate-test-database-by-necessary-data) with some suggestion. Some options are "fixtures" (django's built-in thing) and "factoryboy." @TangoYankee  mentioned the python package [faker](https://faker.readthedocs.io/en/master/) to create fake test data.

I suggest we hash this out on this branch and in the PR comments and merge when ready.

woeip/apps/core/models.py
```diff
@@ -1,6 +1,123 @@
 from django.contrib.auth.models import AbstractUser
+from django.contrib.gis.db.models import LineStringField, PointField
+from django.contrib.postgres.fields import JSONField
+from django.db import models
 
 
 class User(AbstractUser):
     class Meta:
         get_latest_by = 'date_joined'
+
```
clintonb: Consider moving these into a separate app to avoid having too many models in one app. This keeps the code legible.

clintonb (changes requested):
Migrations are missing.

r-b-g-b:
> What error do you get? If you don’t have a virtualenv, open a shell into the container and run the commands there. See the Makefile for the shell command.

Sorry, I was saying that on travis we get these [errors](https://travis-ci.org/openoakland/woeip/builds/468771350#L1275)

I have experienced this locally when I run the app before doing the db migration. I'm guessing that step is not happening during the travis build and was wondering what the best practice is for getting that to happen.

r-b-g-b:
I think the problem is that we don't run ``makemigrations`` anywhere. There is a make recipe for ``detect_missing_migrations``. Perhaps we need to call that in the ``docker-entrypoint.sh``.

r-b-g-b:
Ah I think I understand this error now. I should have committed the migration file, right?

woeip/apps/core/models.py
r-b-g-b: I implemented ``__str__`` on any class that I wanted to interact with on the admin site. The default is the class name and primary key, not v informative. Sure, I might have gone overboard on this one.

woeip/apps/core/models.py
r-b-g-b: Are you suggesting give the ``User`` model in its own app ``core`` and move everything else to a new app?

woeip/apps/core/models.py
r-b-g-b: Just to make sure I'm following the right docs, this? https://docs.djangoproject.com/en/dev/topics/auth/default/#permissions-and-authorization

Would you then recommend replacing the ``Participant`` model with the existing ``User`` model?

r-b-g-b: https://docs.djangoproject.com/en/2.1/topics/auth/customizing/#extending-the-existing-user-model The ``OneToOneField`` option might be a good way to achieve what we want.

From their example:
```
from django.contrib.auth.models import User

class Employee(models.Model):
    user = models.OneToOneField(User, on_delete=models.CASCADE)
    department = models.CharField(max_length=100)
```

.travis.yml
```diff
@@ -13,6 +13,7 @@ install:
   - make docker.build
 
 script:
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "make migrate"
```
r-b-g-b: Previously build failed, I believe because we weren't applying the migrations. This line added to migrate before other steps. But not we are getting a separate travis build error: https://travis-ci.org/openoakland/woeip/builds/468789778#L991

Seems like the ``app`` container is trying to connect to the ``db`` container before the db is finished initializing. . This is happening even with the docker-compose setting ``app: { depends_on: [ db ]}``. I think the problem is ``depends_on`` only checks that the container has started -- there are additional setup steps that the ``db`` container needs to run before ``app`` can contact it without crashing.

r-b-g-b (commented):
Better strategy might be to implement something like this https://docs.docker.com/compose/startup-order/

``app`` tests whether or not the database is up before proceeding.

.travis.yml
r-b-g-b: https://docs.docker.com/compose/startup-order/

clintonb:
Fixtures will grow stale. I prefer factorboy and faker. 

Your models aren’t complex, so you only need to test the `str` method.

woeip/apps/core/models.py
clintonb: What is the purpose of this model?

woeip/apps/core/models.py
clintonb: I’m suggesting moving the new models to their own app.

.travis.yml
clintonb: You will probably want to modify the entry point script to (a) wait for Postgres to be ready and (b) support passing in the script commands. I can take this on sometime tomorrow.

woeip/apps/core/models.py
r-b-g-b: The purpose of ``Participant`` is to hold info about individual WOEIP citizen scientists (or other participants in the future). That way we can link them to datasets they collected/uploaded/etc. We also want to be able to limit read/write access to files, so some way of implementing user/group ownership. Storing contact info is nice, but could be nullable.

woeip/apps/database/models.py
```diff
@@ -0,0 +1,118 @@
...
+    email = models.CharField(max_length=256)
+    contact = models.ForeignKey('Participant',
+                                related_name='contact', related_query_name='contact',
+                                on_delete=models.SET_NULL, blank=True, null=True)
+
+    def __str__(self):
+        return f"{self.name} <{self.email}>"
+
+
+class RoleEnum(models.Model):
```
clintonb: What is the purpose of this model?

woeip/apps/database/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class DatabaseConfig(AppConfig):
```
clintonb: Apps typically have a less-generic name, such as `data_collection`.

woeip/apps/database/models.py
```diff
@@ -0,0 +1,118 @@
...
+    email = models.CharField(max_length=256)
+    organization = models.ForeignKey(Organization, on_delete=models.SET_NULL, blank=True, null=True)
+    role = models.ForeignKey(RoleEnum, on_delete=models.SET_NULL, blank=True, null=True)
+
+    def __str__(self):
+        return f"{self.name} <{self.email}> {self.organization} {self.role}"
+
+
+class RouteEnum(models.Model):
+    """Named routes
```
clintonb: One line.

woeip/apps/database/models.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    name = models.CharField(max_length=256)
+    latlon = LineStringField()
```
clintonb: Routes typically have multiple points. What is this model supposed to represent?

woeip/apps/database/models.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __str__(self):
+        return f"{self.name}"
```
clintonb: No need for `f`.

woeip/apps/database/models.py
```diff
@@ -0,0 +1,118 @@
...
+    def __str__(self):
+        return f"{self.name} {self.model_number} {self.serial_number}"
+
+
+class Sensor(models.Model):
+    """A sensor is something that measures something, i.e., it produces a single measurement value
+    at a time.
+    """
+    name = models.CharField(max_length=256)
+    unit = models.CharField(max_length=256)
```
clintonb: What is a unit? Consider using `help_text` to clarify some of these fields.

woeip/apps/database/models.py
```diff
@@ -0,0 +1,118 @@
...
+    device = models.ForeignKey(Device, on_delete=models.CASCADE)
+
+    def __str__(self):
+        return f"{self.name} ({self.device.name})"
+
+
+class SessionData(models.Model):
+    """The raw data file generated during a session. Assumes one and only one file per sensor,
+    although multiple sensors can be linked to one session"""
+    uri = models.CharField(max_length=256)
```
clintonb: What is this?

woeip/apps/database/models.py
```diff
@@ -0,0 +1,118 @@
...
+            crud.append('update')
+        if self.can_delete:
+            crud.append('delete')
+
+        crud = '/'.join(crud)
+
+        return f"{self.name} ({crud})"
+
+
+class Participant(models.Model):
```
clintonb: This feels a lot like `User`.

woeip/apps/database/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
clintonb: No need to commit empty files.

woeip/apps/database/models.py
r-b-g-b: ``RoleEnum`` will probably get removed once I figure out how to implement permissions using django's built in functionality.

woeip/apps/database/models.py
r-b-g-b: I'm fine with rolling this into ``User``.

woeip/apps/database/models.py
r-b-g-b: ``LineString`` is exactly that. https://shapely.readthedocs.io/en/latest/manual.html#linestrings

woeip/apps/database/models.py
r-b-g-b: By "this" do you mean the ``SessionData`` model? ``SessionData`` is the bridge between unprocessed raw data files and the preprocessed ``Data`` table. Nesting it this way means we can make ``Data`` have as few fields as possible -- the value, time, location, and a link to the session it came from. I considered it important to keep the schema for ``Data`` lean since it will have many rows (several thousand rows per session).

clintonb: I meant the line on which I commented, the `uri` field. What is `uri` pointing to?

woeip/apps/database/models.py
clintonb: There is a concept called "You Ain't Gonna Need It," abbreviated as YAGNI. The odds are some of the code we think we will need later probably won't be used. So, if you know this will be removed later, don't bother adding it now.

clintonb:
https://github.com/openoakland/woeip/pull/12 will fix the build issues. Once it's merged, rebase your branch on master.

This build combines the two branches: https://travis-ci.org/openoakland/woeip/builds/469059472

.travis.yml
clintonb: This can go away.

woeip/apps/data_collection/models.py
```diff
@@ -0,0 +1,118 @@
...
+            crud.append('update')
+        if self.can_delete:
+            crud.append('delete')
+
+        crud = '/'.join(crud)
+
+        return f"{self.name} ({crud})"
+
+
+class Participant(models.Model):
```
clintonb: It seems the only purpose of this model is to link people to organizations. Consider using `User` as your people model and adding an m2m field to that model to link each user to one, or more, `Organization` instances. See https://docs.djangoproject.com/en/2.1/topics/db/examples/many_to_many/ for more details. Note that this will necessitate moving `Organization` into the `core` app.

woeip/apps/data_collection/models.py
```diff
@@ -0,0 +1,118 @@
...
+    name = models.CharField(max_length=256)
+    email = models.CharField(max_length=256)
+    organization = models.ForeignKey(Organization, on_delete=models.SET_NULL, blank=True, null=True)
+    role = models.ForeignKey(RoleEnum, on_delete=models.SET_NULL, blank=True, null=True)
+
+    def __str__(self):
+        return f"{self.name} <{self.email}> {self.organization} {self.role}"
+
+
+class RouteEnum(models.Model):
```
clintonb: `Enum` generally implies a fixed set of entries. Consider the name `Route`.

woeip/apps/data_collection/models.py
```diff
@@ -0,0 +1,118 @@
...
+    """Named routes"""
+    name = models.CharField(max_length=256)
+    latlon = LineStringField()
```
clintonb: Thanks for clarifying the field type. Given the value is more than a single point, the name `latlon` is confusing. Consider a name that implies multiple points (e.g. `path`).

woeip/apps/data_collection/models.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __str__(self):
+        return self.name
+
+
+class Session(models.Model):
+    """A single air quality outing. Can link to several SessionData, e.g., raw data files."""
+    date_collected = models.DateTimeField(auto_now_add=True)
+    route = models.ForeignKey(RouteEnum, on_delete=models.SET_NULL, blank=True, null=True)
+    collected_by = models.ForeignKey(Participant, on_delete=models.CASCADE)
```
clintonb: You probably don't want `CASCADE` for any of your models. You want to make it as difficult as possible to accidentally delete data.

woeip/apps/data_collection/models.py
```diff
@@ -0,0 +1,118 @@
...
+        return f"{self.name} {self.model_number} {self.serial_number}"
+
+
+class Sensor(models.Model):
+    """A sensor is something that measures something, i.e., it produces a single measurement value
+    at a time.
+    """
+    name = models.CharField(max_length=256)
+    unit = models.CharField(max_length=256,
+                            help_text="""Measurement unit, e.g., mg/m3, ppm, etc.""")
```
clintonb: You only need one set of double-quotes, not three.

woeip/apps/database/models.py
r-b-g-b: Ah ``uri`` points to the raw data file, e.g., s3://big-ol-data-file.csv. We decided that it was important to be able to track raw data alongside the processed data in the ``Data`` model.

woeip/apps/data_collection/models.py
```diff
@@ -0,0 +1,118 @@
...
+    device = models.ForeignKey(Device, on_delete=models.CASCADE)
+
+    def __str__(self):
+        return f"{self.name} ({self.device.name})"
+
+
+class SessionData(models.Model):
+    """The raw data file generated during a session. Assumes one and only one file per sensor,
+    although multiple sensors can be linked to one session"""
+    uri = models.CharField(max_length=256)
```
clintonb: Django handles file uploads for you. There is no need to have an explicit URI field. See https://simpleisbetterthancomplex.com/tutorial/2017/08/01/how-to-setup-amazon-s3-in-a-django-project.html and https://docs.djangoproject.com/en/2.1/topics/files/. I recommend adding this functionality in a separate PR to keep things simple in this one.

woeip/apps/data_collection/models.py
TangoYankee: The ```enum``` threw me off in both ```RouteEnum``` and ```RoleEnum``` I would prefer ```Route``` and ```Role```

woeip/apps/data_collection/models.py
```diff
@@ -0,0 +1,118 @@
...
+    email = models.CharField(max_length=256)
+    contact = models.ForeignKey('Participant',
+                                related_name='contact', related_query_name='contact',
+                                on_delete=models.SET_NULL, blank=True, null=True)
+
+    def __str__(self):
+        return f"{self.name} <{self.email}>"
+
+
+class RoleEnum(models.Model):
```
TangoYankee: My understanding is that we decided against defined roles. Database permissions would be handled through the ```Organization``` model

woeip/apps/data_collection/models.py
TangoYankee: A one-to-many relationship would be fine. I do like the idea of using a pre-built ```User``` Model.

woeip/apps/data_collection/models.py
```diff
@@ -0,0 +1,118 @@
...
+        crud = '/'.join(crud)
+
+        return f"{self.name} ({crud})"
+
+
+class Participant(models.Model):
+    name = models.CharField(max_length=256)
+    email = models.CharField(max_length=256)
+    organization = models.ForeignKey(Organization, on_delete=models.SET_NULL, blank=True, null=True)
+    role = models.ForeignKey(RoleEnum, on_delete=models.SET_NULL, blank=True, null=True)
```
TangoYankee: I'm still not convinced of creating roles.

woeip/apps/data_collection/admin.py
```diff
@@ -0,0 +1,15 @@
+from django.contrib import admin
```
TangoYankee: ```data_collection``` is a really broad name for an application. Could we make it more specific, i.e) ```air_quality```

woeip/apps/database/models.py
TangoYankee: "You Ain't Gonna Need it" resonates with me more than "Tables are cheap"

r-b-g-b:
We decided to move forward by simplifying the user/permissions part of the database for now. That involves removing the `Organization`, `Participant`, and `Role` tables. For now, all users will have permission for all actions.

r-b-g-b:
Rethinking my previous comment, it is useful to keep track of participants, e.g., I want to retrieve all the data collected by a particular participant. I propose keeping the `Participant` table (and `Organization` table) but not implementing any participant-level permissions.

.travis.yml
TangoYankee: ...."This can go away" sounds like a mobster threat

woeip/apps/data_collection/models.py
TangoYankee: Should we have a docker for the S3 bucket, too? We were looking at a [Storage Container](https://docs.docker.com/registry/storage-drivers/s3/#parameters)

clintonb:
Use `User` rather than `Participnt`.

woeip/apps/data_collection/models.py
clintonb: Nope. Data volumes are used locally.

r-b-g-b:
I'm wondering about that. The easiest thing would be to have a `Participant` table, just holding info about participants, like any other kind of non-person data. And use `User` to do actual authentication.

clintonb:
Keep it simple. If you don’t have a great need/story around permissions, use `User` and modify it later. As it is, you’re honestly doing a bit too much for one PR. I recommend picking one important aspect (e.g. sensor data collection) and perfecting that. Come back and update permissions with a second PR.

woeip/apps/data_collection/models.py
r-b-g-b: Agreed that `CASCADE` should be used rarely. Two places I think it's appropriate: the links between `Session` -< `SessionData` -< `Data`.

- `Session` contains info about the data collection outing: date collected, route, collected_by, and session metadata
- `SessionData` contains info about the raw data file: which device produced it (and the measurement units), upload time, uploaded_by
- `Data` contains the individual data points of the raw data broken out into individual rows.

I think when a user deletes a session, they expect the `CASCADE` behavior: that the raw data and individual measurements will be deleted too.

r-b-g-b:
I think that's what I'm suggesting, but I could be misunderstanding what you're saying. How about this -- for this earliest stage: no `User` model at all. Anyone who wants to add to/delete from the database gets a superuser account through django. `Participant` is kept as-is to keep track of who collected what.

woeip/apps/data_collection/models.py
r-b-g-b: This is cool. Can you say more about how we simulate this locally using data volumes? I don't understand what that will look like.

clintonb:
The `User` model already exists, because Django needs a user model. Superuser means something very specific in Django terms. You do _not_ want everyone to be a superuser. You can give all users the ability to modify all data by omitting any restrictions on those actions.

I would do the following:
1. Remove `Participant` and `Organization` for the time being, and focus solely on `Session`, `SessionData`, and related models.
2. Set your people-related foreign key fields to point to `core.User`.

r-b-g-b:
Ok, I'm gonna do this now and think about why it works later :)

woeip/apps/data_collection/models.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __str__(self):
+        return f"{self.name} <{self.email}> {self.organization} {self.role}"
+
+
+class RouteEnum(models.Model):
+    """Named routes"""
+    name = models.CharField(max_length=256)
```
clintonb: Does this field need to be unique?

woeip/apps/data_collection/models.py
```diff
@@ -0,0 +1,118 @@
...
+    def __str__(self):
+        return self.name
+
+
+class Session(models.Model):
+    """A single air quality outing. Can link to several SessionData, e.g., raw data files."""
+    date_collected = models.DateTimeField(auto_now_add=True)
+    route = models.ForeignKey(RouteEnum, on_delete=models.SET_NULL, blank=True, null=True)
+    collected_by = models.ForeignKey(Participant, on_delete=models.CASCADE)
+    metadata = JSONField(blank=True, null=True)
```
clintonb: What metadata do you expect to go in this field? Can/should those fields be on the model itself? My YAGNI sense is going off here.

woeip/apps/data_collection/models.py
```diff
@@ -0,0 +1,118 @@
...
+    def __str__(self):
+        return f"{self.name} {self.model_number} {self.serial_number}"
+
+
+class Sensor(models.Model):
+    """A sensor is something that measures something, i.e., it produces a single measurement value
+    at a time.
+    """
+    name = models.CharField(max_length=256)
+    unit = models.CharField(max_length=256,
```
clintonb: How many different units do you expect? If the number is limited, it might make sense to use [`choices`](https://docs.djangoproject.com/en/2.1/ref/models/fields/#choices) here.

woeip/apps/data_collection/models.py
```diff
@@ -0,0 +1,118 @@
...
+
+class SessionData(models.Model):
+    """The raw data file generated during a session. Assumes one and only one file per sensor,
+    although multiple sensors can be linked to one session"""
+    uri = models.CharField(max_length=256)
+    sensor = models.ForeignKey(Sensor, on_delete=models.SET_NULL, blank=True, null=True)
+    session = models.ForeignKey(Session, on_delete=models.CASCADE)
+    upload_time = models.DateTimeField(auto_now_add=True)
+    uploaded_by = models.ForeignKey(Participant, on_delete=models.SET_NULL, blank=True, null=True)
+    hash = models.CharField(max_length=256)
```
clintonb: I'm assuming this is supposed to represent a computed hash. What are you hashing to generate this?

r-b-g-b: I was thinking of this as a way to ensure that the same raw data file isn't uploaded twice. So hashing the content of the raw data file itself.

woeip/apps/data_collection/models.py
r-b-g-b: Eh, I'm no expert, but I've seen dozens of units for different types of sensors. With `choices`, how easy would it be to add units in the future if we forgot a few?

woeip/apps/data_collection/models.py
clintonb: The simple version is Django is configured to store data in a directory local to the Docker container. The complicated version is that this directory is mapped to a data volume stored on the host machine. More info on Docker volumes is available at https://docs.docker.com/storage/volumes/.

Django's `FileField` can manage the storage for you, whether it be local storage on the Docker volume or (eventually) storage on S3. You don't need an explicit URI field. You instead need a `FileField`: `upload = models.FileField(upload_to='session_data')`.

clintonb:
@r-b-g-b I realize I keep dropping comments in waves. Let me know if I'm becoming overbearing, and you need some relief. I'm also happy to schedule some time to talk through this synchronously.

woeip/apps/data_collection/models.py
clintonb: Adding a new choice is as simple as adding an item to an array.

woeip/apps/data_collection/models.py
clintonb: Got it. I have a bunch of thoughts on this:

1. MD5 hashes are only 32 characters, so you can decrease `max_length`.
2. There is some [prior art](https://stackoverflow.com/questions/38256201/django-prevent-file-duplicates-in-view) you can use to override the `save` method to do this.
3. I recommend doing this work in a follow-on PR.

r-b-g-b:
Thanks for checking. Not overbearing at all, I'm finding it extremely helpful! That said, you might be onto something that a few minute chat could clear things up and get this thing finished. I'm free Fri morning or most times over the weekend.

clintonb:
@r-b-g-b I sent you a link via Slack for scheduling. I'm two time zones ahead, and headed to bed.

r-b-g-b:
Pushing a bunch of changes that address stuff we've discussed in the 74 (!!) comments so far. Only a few things left to cover:
- replace `SessionData.uri` with built-in django storage functionality
- use `choices` for `Sensor.units`

.travis.yml
```diff
@@ -14,6 +14,7 @@ install:
...
```
clintonb: This line still needs to be removed.

woeip/apps/air_quality/factories.py
```diff
@@ -0,0 +1,76 @@
...
+from django.contrib.gis import geos
+
+from woeip.apps.air_quality.models import Data, Device, Route, Sensor, Session, SessionData
+from woeip.apps.core.factories import UserFactory
+
+
+def generate_route(n):
+    points = []
+    for _ in range(n):
+        points.append(geos.Point(random.random(), random.random()))
```
clintonb: Those numbers may not be valid coordinates. Try `geos.Point(fake.latlng())`. See https://faker.readthedocs.io/en/master/providers/faker.providers.geo.html for reference.

Alternatively, there is this: https://stackoverflow.com/a/32831123/592820

woeip/apps/air_quality/models.py
```diff
@@ -0,0 +1,73 @@
+from django.contrib.gis.db.models import LineStringField, PointField
+from django.contrib.postgres.fields import JSONField
+from django.db import models
+
+from woeip.apps.core.models import User
+
+
+class Route(models.Model):
+    """Named routes"""
```
clintonb: This docstring is no longer needed.

woeip/apps/air_quality/models.py
```diff
@@ -0,0 +1,73 @@
...
+    def __str__(self):
+        return f"{self.name} {self.model_number} {self.serial_number}"
+
+
+class Sensor(models.Model):
+    """A sensor is something that measures something, i.e., it produces a single measurement value
+    at a time.
+    """
+    name = models.CharField(max_length=256)
+    unit = models.CharField(max_length=256,
```
clintonb, line 35: I opened https://github.com/openoakland/woeip/issues/13 to modify this later.

woeip/apps/air_quality/models.py
```diff
@@ -0,0 +1,73 @@
...
+    def __str__(self):
+        return f"{self.name} ({self.device.name})"
+
+
+class Session(models.Model):
+    """A single air quality outing. Can link to several SessionData, e.g., raw data files."""
+    date_collected = models.DateTimeField(auto_now_add=True)
+    route = models.ForeignKey(Route, on_delete=models.SET_NULL, blank=True, null=True)
+    collected_by = models.ForeignKey(User, on_delete=models.SET_NULL, blank=True, null=True)
+    metadata = JSONField(blank=True, null=True)
```
clintonb: I suggest removing this field for now. Add it later when there is a great need for it. If you add it now, you run the risk of it being an unstructured catch-all field that will become a mess to deal with.

woeip/apps/air_quality/models.py
```diff
@@ -0,0 +1,73 @@
...
+
+class SessionData(models.Model):
+    """The raw data file generated during a session. Assumes one and only one file per sensor,
+    although multiple sensors can be linked to one session"""
+    uri = models.CharField(max_length=256)
+    sensor = models.ForeignKey(Sensor, on_delete=models.SET_NULL, blank=True, null=True)
+    session = models.ForeignKey(Session, on_delete=models.CASCADE)
+    upload_time = models.DateTimeField(auto_now_add=True)
+    uploaded_by = models.ForeignKey(User, on_delete=models.SET_NULL, blank=True, null=True)
+    hash = models.CharField(max_length=256)
```
clintonb: I suggest removing this field, and adding a compound unique key on the `sensor` and `session` fields. I would guess the uniqueness would prevent most of the duplicate upload issues. See https://docs.djangoproject.com/en/2.1/ref/models/options/#unique-together.

woeip/apps/air_quality/models.py
```diff
@@ -0,0 +1,73 @@
...
+    metadata = JSONField(blank=True, null=True)
+
+    def __str__(self):
+        return f"{self.date_collected} {self.collected_by}"
+
+
+class SessionData(models.Model):
+    """The raw data file generated during a session. Assumes one and only one file per sensor,
+    although multiple sensors can be linked to one session"""
+    uri = models.CharField(max_length=256)
```
clintonb: Change this to `upload = models.FileField(upload_to='session_data')`. See the previous comments for why.

woeip/apps/core/models.py
```diff
@@ -1,6 +1,12 @@
 from django.contrib.auth.models import AbstractUser
+from django.db import models
 
 
 class User(AbstractUser):
     class Meta:
         get_latest_by = 'date_joined'
+
+    email = models.CharField(max_length=256)
```
clintonb: `AbstractUser` already has an `email` field. This file does not need to be modified at this time.

clintonb (changes requested):

woeip/apps/core/factories.py
```diff
@@ -0,0 +1,11 @@
+import factory
+
+from woeip.apps.core.models import User
+
+
+class UserFactory(factory.DjangoModelFactory):
+    class Meta:
+        model = User
+
+    name = factory.Faker('name')
```
clintonb: `User` doesn't have a `name` field. See https://github.com/clintonb/cookiecutter-django/pull/29/files for a complete factory.

woeip/apps/air_quality/factories.py
```diff
@@ -0,0 +1,74 @@
...
+import factory.fuzzy
+from django.contrib.gis import geos
+
+from woeip.apps.air_quality.models import Data, Device, Route, Sensor, Session, SessionData
+from woeip.apps.core.factories import UserFactory
+
+
+class FuzzyLineString(factory.fuzzy.BaseFuzzyAttribute):
+    def fuzz(self):
+        points = [(random.random(), random.random())
```
clintonb: Use `factory.Faker('latlng')` to generate real coordinates. See https://faker.readthedocs.io/en/master/providers/faker.providers.geo.html for other providers.

woeip/apps/air_quality/models.py
```diff
@@ -0,0 +1,75 @@
...
+    name = models.CharField(max_length=256)
+    unit = models.CharField(max_length=256,
+                            help_text="Measurement unit, e.g., mg/m3, ppm, etc.")
+    device = models.ForeignKey(Device, on_delete=models.SET_NULL, blank=True, null=True)
+
+    def __str__(self):
+        return f"{self.name} ({self.device.name})"
+
+
+class Session(models.Model):
```
clintonb, line 43: Consider subclassing [`TimeStampedModel`](https://django-extensions.readthedocs.io/en/latest/model_extensions.html?highlight=timestampedmodel) to get `created` and `updated` fields on this and other models.

woeip/apps/core/migrations/0001_initial.py
```diff
@@ -1,5 +1,7 @@
-# Generated by Django 2.1.3 on 2018-11-19 07:45
```
clintonb: There shouldn't be any changes to this file.

woeip/apps/core/factories.py
```diff
@@ -0,0 +1,19 @@
+import factory
```
clintonb: Since factories are test-related I tend to put them in `tests` directories (e.g. `woeip/apps/core/tests/factories.py`).

clintonb (commented):
This is nearly ready to merge. I left a few comments. Once they are addressed I 👍.

woeip/apps/air_quality/factories.py
TangoYankee: Is it possible to bound it to within Oakland?

clintonb: Yes, but not really worth the hassle. The models don't have any location constraints, so neither should the test data.

r-b-g-b: Ehhh I looked into that and was not convinced it was very useful. Faker's latlng just samples from a list of world city lat/longs, which is just as meaningless as random points when it comes to making routes. If we come up with a test that relies on having realistic routes, I'd be down to change it, but for now I'd just say leave it as is.

woeip/apps/core/migrations/0001_initial.py
r-b-g-b: I think I changed this to avoid make quality failures. I could be wrong. isort definitely complains about that import order, but should migrations files be ignored?

clintonb: Migrations are not ignored since they are code and can be manually modified.

No changes are needed to this file. You added imports, but they are not used.

r-b-g-b: Ah that's right!

woeip/apps/air_quality/admin.py
```diff
@@ -0,0 +1,15 @@
...
+
+class SessionDataAdmin(admin.ModelAdmin):
+    readonly_fields = ('created', 'modified', )
+
+
+admin.site.register(Route)
+admin.site.register(Session)
+admin.site.register(Device)
+admin.site.register(Sensor)
+admin.site.register(SessionData, SessionDataAdmin)
```
clintonb: Note that you can use a decorator:

```python
@admin.register(SessionData)
class SessionDataAdmin(admin.ModelAdmin):
  readonly_fields = ('created', 'modified', )
```

clintonb (approved):

clintonb:
P.S. There's probably not a good reason to maintain 25 commits of history. In such cases I either squash locally and force-push or use the "Squash and merge" option to merge.

r-b-g-b:
Yay! Thanks for all the work on this one everyone!

Oops on that commit message -- obv not WIP anymore. I tried amending the message, but I don't have permission to force push. Oh well!

TangoYankee:
> Fixtures will grow stale. I prefer factorboy and faker.
> 
> Your models aren’t complex, so you only need to test the `str` method.

@clintonb I'm inexperienced with testing and not sure what level of complexity calls for testing. Could you point me to some good resources to learn?

clintonb:
@TangoYankee that's a good question to which I cannot point to any great resources. I learned a lot via trial-and-error and mentorship. I'll try to find some content.

My personal philosophy is to test the bits of the code that I have customized rather than the framework I have customized. In this case, adding a model doesn't warrant testing because all we've done is utilize the framework (which is already tested). We did customize `str`, so that can be tested. If we added new methods to the models, we would want to test those new methods.

As we start adding views, especially API views that return data, we'll want to test those responses because we get away from the framework and more toward the business logic.

TangoYankee:
I appreciate it

PR #12, opened 2018-12-17
Updated Travis scripts

clintonb:
- We now pull images prior to building. This helps avoid wasting time completely rebuilding the app container.
- Added test to ensure migrations have been generated, and run properly.
- We wait for the database to become available before running Django management commands

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,15 @@ cache: pip
 
 install:
   - bash .travis/install_terraform.sh
+  - make docker.pull
   - make docker.build
 
 script:
   - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "make quality"
-  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "make static"
-  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "make test"
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "./wait-for-it.sh db:5432 -- make detect_missing_migrations"
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "./wait-for-it.sh db:5432 -- make migrate"
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "./wait-for-it.sh db:5432 -- make static"
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "./wait-for-it.sh db:5432 -- make test"
   - make validate_terraform
 
 after_success:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -29,6 +29,9 @@ clean: ## Delete generated byte code and coverage reports
 	find . -name '*.pyc' -delete
 	coverage erase
 
+detect_missing_migrations:  ## Determine if any apps are missing generated migrations
```
r-b-g-b: This exists already, no? https://github.com/openoakland/woeip/blob/82e36bf5c4b0806c5fd32dd3a5dcc0844aee9d3d/Makefile#L55

.travis.yml
```diff
@@ -10,11 +10,15 @@ cache: pip
 
 install:
   - bash .travis/install_terraform.sh
+  - make docker.pull
   - make docker.build
 
 script:
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml up -d db
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml run app "make detect_missing_migrations"
```
r-b-g-b: Is there any hard check that the postgis service is initialized and ready to accept connections, or does this just hope that doing ``up -d db`` first will give it enough time to start before the ``detect_missing_migrations``.

https://travis-ci.org/openoakland/woeip/builds/469069161#L1022

Seems to need a hard check solution:
- [wait-for-it](https://github.com/vishnubob/wait-for-it/blob/master/wait-for-it.sh)
- [good discussion here](https://github.com/docker-library/postgres/issues/146)

TangoYankee:
What do you think is causing: ```psycopg2.OperationalError: could not connect to server: Connection refused```?

clintonb:
The database container hasn’t started. I thought I’d resolved this.

TangoYankee:
Yeah, my understanding is still too limited to make a meaningful contribution

clintonb:
This is ready. @TangoYankee or @r-b-g-b, please approve so that I can merge.

r-b-g-b:
Nice! Looks like it's behaving better now. One issue I read about involves wait-for-it.sh and the way the postgres docker containers are instantiated. wait-for-it.sh just checks that a given host:port responds. While initializing itself, the postgres container starts and stops several times before it is actually ready. It is possible that wait-for-it.sh falsely detects the postgres service during one its intermediate phases. No sense for how likely this is to ever occur, just something to keep in mind if we ever notice random failed tests. I think this looks ready to merge, thanks @clintonb !

r-b-g-b (approved):

clintonb:
@r-b-g-b your concern is valid in general, but not an issue in this specific case. The restart issues only apply on the first run of the database. The first line of `script` starts both containers. We don't actually need database access until the calls to the `migrate` and `make` targets. The Postgres container initialization is relatively quick, and is completed by the time we reach those lines in the script.

I imagine this race condition will be fine for the foreseeable future. If any drastic changes are made to that container, we should definitely revisit this issue.

PR #14, opened 2019-01-09
Refactor joiner functionality

r-b-g-b:


Ethan-bradley (commented):
This code is a revised version of the former joiner.py. It combines the air quality and gps files and then in the second commit file sends them to the database. This code creates an air quality file that almost exactly matches the original, the differences being a minor problem involving one data point.

r-b-g-b:
Thanks Ethan, that's a good description of what this is about. I won't make a habit of comment-less PRs, but wanted to put the code somewhere so we could look at it at the meetup tonight.

The discrepancy Ethan refers to is a slight difference in the algorithm to link GPS and air quality measurements. The original only looked for GPS points after the air quality measurement, this one will take the closest point even if it earlier in time.

woeip/apps/air_quality/dustrak.py
```diff
@@ -0,0 +1,295 @@
...
+import pandas as pd
+import pynmea2
+import pytz
+from django.contrib.gis import geos
+
+from . import models
+
+
+def parse_header(f):
+    """Parse the header for a DusTrak II file.
```
clintonb, line 16: You are using the [NumPy docstring style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html#example-numpy). The [Google docstring style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). Whichever you pick, just make sure folks stick with it throughout the codebase.

woeip/apps/air_quality/dustrak.py
```diff
@@ -0,0 +1,295 @@
...
+    parser.add_argument('-t', '--tz', type=str, default="America/Los_Angeles",
+                        help="Time zone, e.g., America/Los_Angeles")
+    parser.add_argument('--tolerance', type=float, default=3.,
+                        help="How far away (in seconds) can a air quality measurement be from a "
+                             "GPS measurement to be linked to it")
+
+    return parser.parse_args()
+
+
+def main():
```
clintonb: Looking higher level, how is this bit of code supposed to be used? I assume it will be called shortly after files are uploaded. If that's the case, you don't need a main function as this isn't going to be a script called from the command line. Similarly, you aren't going to accept file paths as an input.

Consider updating `join()` to accept the raw files rather than their paths. This will allow you to call `join()` from your code responsible for handling the uploads.

If you have a need to call this code from the command line, create a [management command](https://docs.djangoproject.com/en/2.1/howto/custom-management-commands/).

woeip/apps/air_quality/dustrak.py
```diff
@@ -0,0 +1,295 @@
...
+        else:
+            longitudes.append(-lon)
+
+    gps['lat'] = latitudes
+    gps['lon'] = longitudes
+
+    return gps
+
+
+def join(dustrak_path, gps_path, tz='America/Los_Angeles', tolerance=3.):
```
clintonb: This definitely needs testing. At a minimum, you should pass in data and ensure the joined data is written to the database as expected.

woeip/apps/air_quality/dustrak.py
```diff
@@ -0,0 +1,295 @@
...
+    The DusTrak device collects time stamps and air quality measurements, but no geospatial
+    information. Scientists will also take a GPS device on their sessions that records time stamps
+    and latitudes/longitudes. Those two data files need to be combined.
+
+    Parameters
+    ----------
+    dustrak_path : str
+    gps_path : str
+    tz : str
+    tolerance : numeric
```
clintonb, line 228: s/numeric/float

\* `s/` is the command for search and replace in vi and other Unix tools: http://www.linfo.org/vi/search.html.

woeip/apps/air_quality/dustrak.py
```diff
@@ -0,0 +1,295 @@
...
+
+        warnings.warn(message)
+
+    joined_data = joined_data[['time', 'Mass [mg/m3]', 'lat', 'lon']]
+    joined_data.rename(columns={'Mass [mg/m3]': 'measurement'}, inplace=True)
+
+    return joined_data
+
+
+def save(joined_data, session_data):
```
clintonb, line 257: This is never called. Dead code should not be in the code base. Remove it, and restore it when it is actually in use.

woeip/apps/air_quality/dustrak.py
```diff
@@ -0,0 +1,295 @@
...
+    ----------
+    joined_data : pandas DataFrame
+    session_data : woeip.apps.air_quality.models.SessionData
+    """
+    for _, row in joined_data.iterrows():
+        data = models.Data(session_data=session_data,
+                           value=row['measurement'],
+                           time=row['time'],
+                           latlon=geos.Point(row['lon'], row['lat']))
+        data.save()
```
clintonb: Use [`create()`](https://docs.djangoproject.com/en/2.0/ref/models/querysets/#create) to instantiate and save the object to the database in one call:

```python
models.Data.objects.create(
  session_data=session_data,
  value=row['measurement'],
  time=row['time'],
  latlon=geos.Point(row['lon'], row['lat'])
)
```

woeip/apps/air_quality/dustrak.py
```diff
@@ -0,0 +1,295 @@
...
+
+def save(joined_data, session_data):
+    """Save a table of joined data to the database
+
+    Parameters
+    ----------
+    joined_data : pandas DataFrame
+    session_data : woeip.apps.air_quality.models.SessionData
+    """
+    for _, row in joined_data.iterrows():
```
clintonb, line 266: Use [`bulk_create()`](https://docs.djangoproject.com/en/2.0/ref/models/querysets/#bulk-create) to create all items in a single database call:

```python
models.Data.objects.bulk_create([
  models.Data(session_data=session_data,
    value=row['measurement'],
    time=row['time'],
    latlon=geos.Point(row['lon'], row['lat']))
  for _, row in joined_data.iterrows()
])
```

woeip/apps/air_quality/dustrak.py
```diff
@@ -0,0 +1,295 @@
...
+    joined_data = pd.merge_asof(data, gps, on='time', direction='nearest',
+                                tolerance=pd.Timedelta(f'{tolerance}s'))
+
+    invalid_indices = joined_data[['lat', 'lon', 'Mass [mg/m3]']].isnull().any(1)
+    joined_data = joined_data[~invalid_indices]
+
+    n_dropped = invalid_indices.sum()
+
+    if n_dropped > 0:
+        message = (f"""{n_dropped} air quality samples dropped that were """
```
clintonb: This definitely fits on a single line: `message = f"{n_dropped} air quality samples dropped that were not within {tolerance} seconds of a GPS sample."`.

clintonb (changes requested):
I noticed a couple minor things:
The PR title is still prefixed with "WIP". Remember to remove this before merging.

Let me know if you have questions or need assistance with testing.

woeip/apps/air_quality/dustrak.py
clintonb: This is an FYI. Use `bulk_create()` for this function, assuming the function is not removed.

woeip/apps/air_quality/dustrak.py
r-b-g-b: Will stick with numpy style unless you think Google docstrings are better!

clintonb: I don't think either is better than the other. Use whichever you all prefer.

woeip/apps/air_quality/dustrak.py
r-b-g-b: This will be useful when we create the data upload forms. For now there's at least a test that uses this.

woeip/apps/air_quality/tests/test_dustrak.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=W0621
```
clintonb: Disable pylint warnings for the line with the issue rather than the entire file.

woeip/apps/air_quality/dustrak.py
TangoYankee: Is it a style preference? Or is it common to apply logic that depends on the format of the docstring?

woeip/apps/air_quality/tests/test_dustrak.py
```diff
@@ -0,0 +1,51 @@
...
+            assert all(target_data[column] == joined_data[column])
+        else:
+            assert np.allclose(target_data[column], joined_data[column])
+
+
+@pytest.mark.django_db
+def test_save(target_data):
+    """Test ability to save a session of joined GPS/air quality measurement data to the database"""
+    session_data = factories.SessionDataFactory()
+    session_data.save()
```
clintonb: This is not needed. Factories generate saved objects.

woeip/apps/air_quality/tests/test_dustrak.py
```diff
@@ -0,0 +1,51 @@
...
+
+    dustrak.save(target_data, session_data)
+
+    assert len(models.Data.objects.all()) == len(target_data)
```
clintonb: You should validate the values saved to the DB.

clintonb (commented):
No need to await my approval.

woeip/apps/air_quality/tests/test_dustrak.py
TangoYankee: This is suppressing warnings for use of the ```target_data``` name. For lines 25 and 44, it makes sense to suppress the warning; pylint fails to recognize they are function calls. However, I am less sure about lines 17-22. There, ```target_data``` is defined as a variable within the ```target_data``` function. Would it be better to rename ```target_data``` in lines 17-22 to ```target_output```, rather than simply suppress the warning? Or does pytest.fixtures rely on the logic of calling the function within itself?

woeip/apps/air_quality/tests/test_dustrak.py
TangoYankee: As in, lines 62-65 of ```tests/factories.py``` save the data, when they are created using ```factory```? Consequently, just delete line 47 of ```tests/test_dustrak.py``` ?

woeip/apps/air_quality/tests/test_dustrak.py
TangoYankee: Rather than simply checking the lengths, crack the files open with a for-loop and assert that each line of ```models.Data.objects.all()``` matches ```target_data``` ?

woeip/apps/air_quality/tests/test_dustrak.py
clintonb: The call to the base class saves the object, not lines 62-65.

Yes, delete this line.

woeip/apps/air_quality/tests/test_dustrak.py
clintonb: The fix here is to not use a fixture. This data is only needed by the one test, so there is no need to make a fixture (which you might share with multiple test modules). Instead, just call the function from your tests.

If you must disable a pylint warning, use the name, not the code (e.g. `redefined-outer-name`) to make it clear what's going on without an extra lookup.

woeip/apps/air_quality/tests/test_dustrak.py
TangoYankee: Edit: used a numpy.allclose, similar to the one used in the joiner test_joiner. Limited it to a check of measurements in target data against values in database storage.

woeip/apps/air_quality/tests/test_dustrak.py
TangoYankee: It took me forever to find a [good list of error names](https://github.com/janjur/readable-pylint-messages/blob/master/README.md). Do you just ...*know* them?

clintonb: Nope. When you run pylint the errors are printed in parentheses.

TangoYankee: Hmm, all I saw were the messages and not the names. Guess I know what to look for next time.

TangoYankee (approved):

PR #15, opened 2019-01-09
Adds S3 storage

nbilenko:
S3 storage added for session data files. Env variables `AWS_S3_ACCESS_KEY_ID` and `AWS_S3_SECRET_ACCESS_KEY` need to be configured in `docker-compose.local.yml` for local testing. Bucket name is set to `'woaq'`.

Tested locally by creating new SessionData objects from the admin site, which uploads new files to S3.


clintonb (commented):
1. Added the new environment variables to `docker-compose.travis.yml`.
2. Consider populating `DEFAULT_FILE_STORAGE` from an environment variable so that developers can easily override it for local development and not have to worry about S3 access.

woeip/settings.py
```diff
@@ -40,6 +41,15 @@
 
 INSTALLED_APPS = DJANGO_APPS + THIRD_PARTY_APPS + LOCAL_APPS
 
+
+if len(env('AWS_S3_ACCESS_KEY_ID')) > 10:
```
clintonb: Is this to disable use of S3? If so, this works, but is not very explicit. Consider the alternative:

1. Allow DEFAULT_FILE_STORAGE to be passed as an environment variable, but default to S3. 
2. Set this value to local file storage for local usage in Docker-Compose.local. 
3. Default the S3 settings to None. 
4. If the file storage is set to S3 and the access variables are not set, raise a configuration error.

clintonb (changes requested):

woeip/settings.py
nbilenko: Thanks for the review! Made changes based on your feedback.

clintonb (approved):

PR #17, opened 2019-02-02
Ty/add joiner

TangoYankee:
Make changes based on @clintonb review of r-b-g-b/add-joiner

### DIFF
--- a/woeip/apps/air_quality/tests/test_dustrak.py
+++ b/woeip/apps/air_quality/tests/test_dustrak.py
@@ -1,4 +1,3 @@
-# pylint: disable=W0621
 from pathlib import Path
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


TangoYankee:
This a request onto r-b-g-b/add_joiner, rather than master. Requesting a code review, to avoid a merge that takes us in the wrong direction.

woeip/apps/air_quality/tests/test_dustrak.py
```diff
@@ -14,15 +13,15 @@
...
+    target_output.set_index('time', inplace=True)
+    target_output = target_output.tz_localize('UTC')
+    target_output.reset_index(inplace=True)
 
-    return target_data
+    return target_output
 
 
-def test_joiner(target_data):
+def test_joiner(target_data):  # pylint: disable=W0621
```
clintonb, line 24: Ditch the fixture. Just call the method where needed.

r-b-g-b: Is there a good time to use fixtures? I thought this was an appropriate place to use a fixture -- to avoid running the same code twice.

TangoYankee: From his comment on the other PR:

> This data is only needed by the one test, so there is no need to make a fixture (which you might share with multiple test modules).

TangoYankee: @clintonb @r-b-g-b If everyone is good with the other two changes (removing the save command and changing the assert in the test_save), can I merge this into the r-b-g-b/add-joiner? We'll have just the one PR. We can remove the fixture from that one.

r-b-g-b: The data is needed by two tests though, ``test_joiner`` and ``test_save``?

r-b-g-b: But yeah all good for merging!

r-b-g-b (approved):
Good to go!

woeip/apps/air_quality/tests/test_dustrak.py
clintonb: Most fixtures live in a separate file and are share by multiple test files. As we see, it’s just a function. Since it isn’t being used elsewhere there is no need to create a fixture.

PR #18, opened 2019-02-02
remove pytest fixture. remove suppression of pylint warnings

TangoYankee:
Here is a version of add-joiner that removes the pytest fixture. It also reorganizes the calls to ```target_data```. ```target_data``` is now set as a global variable based on the function, renamed to ```get_target_data```.

### DIFF
--- a/woeip/apps/air_quality/tests/test_dustrak.py
+++ b/woeip/apps/air_quality/tests/test_dustrak.py
@@ -10,8 +10,7 @@ from woeip.apps.air_quality.tests import factories
 test_data_directory = Path(__file__).parent / 'data'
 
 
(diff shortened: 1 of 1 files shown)


r-b-g-b:
I'd like to understand the objection to this fixture. Should they be avoided at all costs? The added complexity doesn't justify a fixture when it only is used twice?

clintonb:
I have no strong objections. In this case the fixture simply isn’t needed given the limited usage. You can just call the function directly.

r-b-g-b:
Ok, I'm fine with that, just wanted to make sure I wasn't missing something. Merge!

r-b-g-b (approved):

PR #22, opened 2019-02-03
Refactor dustrak load and join functions

r-b-g-b:
A few changes to the dustrak loader and joiner. Most importantly, raise ``ValueError`` when air quality or gps files do not load. Return the error message that the loader raised. Also, open the files in binary mode and catch multiple different kinds of line endings when parsing the header.

### DIFF
--- a/woeip/apps/air_quality/dustrak.py
+++ b/woeip/apps/air_quality/dustrak.py
@@ -127,7 +127,7 @@ def load_dustrak(file_handle, tz):
 
     Parameters
     ----------
-    file_handle : str
+    file_handle : file handle, mode='rb'
 
     Returns
     -------
@@ -167,7 +167,7 @@ def load_gps(file_handle):
 
     Parameters
     ----------
-    file_handle : file stream
+    file_handle : file handle, mode='rb'
 
     Returns
     -------
@@ -223,8 +223,8 @@ def join(dustrak_file_handle, gps_file_handle, tz='America/Los_Angeles', toleran
 
     Parameters
     ----------
-    dustrak_file_handle : file handle
-    gps_file_handle : file handle
+    dustrak_file_handle : file handle, mode='rb'
+    gps_file_handle : file handle, mode='rb'
     tz : str
     tolerance : numeric
         How far away (in seconds) can a air quality measurement be from a GPS measurement to be
--- a/woeip/apps/air_quality/dustrak.py
+++ b/woeip/apps/air_quality/dustrak.py
@@ -26,10 +26,10 @@ def parse_header(f):
     """
     header = {}
 
-    lines = itertools.takewhile(lambda x: x not in (b'\n', b'\r', b'\r\n'), f)
+    lines = itertools.takewhile(lambda x: x not in ('\n', '\r', '\r\n'), f)
     for line in lines:
-        line = line.rstrip(b'\n\r')
-        key, value = line.split(b',')
+        line = line.rstrip('\n\r')
+        key, value = line.split(',')
         header[key] = value
 
     return header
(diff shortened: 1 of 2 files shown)


TangoYankee (changes requested):
Update the doc string to specify that the file is being read in binary mode and why it is done that way

clintonb (changes requested):
Why are you opening a text file in binary mode?

r-b-g-b:
This function is meant to process data files uploaded via the web form. The ``request.FILES['air_quality']`` file opens in binary mode, so this function is made to handle that.

r-b-g-b:
See this for an example for how the function is used: https://github.com/openoakland/woeip/blob/r-b-g-b/upload-form/woeip/apps/air_quality/views.py

clintonb:
I see now. Your files need to be decoded as UTF-8. In general, if you are working with text, work with text. Avoid introducing binary strings to keep encoding/decoding simple. In this case, you need to update https://github.com/openoakland/woeip/blob/r-b-g-b/upload-form/woeip/apps/air_quality/views.py#L40-L41 to decode the files before passing them in.

Try updating these lines to:

```python
from django.utils.encoding import smart_text
...
joined_data = dustrak.join(smart_text(request.FILES['air_quality'].read()),
                           smart_text(request.FILES['gps'].read()))
```

[`smart_text`](https://docs.djangoproject.com/en/2.1/ref/utils/#django.utils.encoding.smart_text) will decode the input files to a proper UTF-8 `str`.

clintonb:
Correction: use `force_text` as recommended by https://docs.djangoproject.com/en/2.1/ref/unicode/#useful-utility-functions.

r-b-g-b:
Whoooa

r-b-g-b:
Ok in that case, this PR is pretty straightforward: catching errors if air quality and gps files are not formatted correctly.

TangoYankee:
@r-b-g-b Before merging this PR, could you create an issue which outlines the conversion of `views` to include `force_text`, to avoid losing track of the task?

Also, delete *open files in binary mode* from PR name?

r-b-g-b:
That is a part of the file upload issue, I'll make note of it there. :+1:

TangoYankee (approved):

r-b-g-b:
@clintonb Does this look ready to merge now?

clintonb (changes requested):
Ditch the file streams/handles, and stick with strings throughout. The quicker you can get to a UTF-8 string, the simpler the code will be.

r-b-g-b:
Yeah, the annoying thing is that there doesn't appear to be a pandas method that loads directly from stream. So it's necessary to wrap any string in a ``io.StringIO`` at some point. I could move that inside the ``dustrak.joiner``, but I figured it was more natural for a loader to accept a file handle-like object than a string containing the file contents.

clintonb:
It's not necessarily more natural to accept a file handle, especially given we aren't explicitly operating on files. If you use `string` at the interface, you can push the file-handling logic solely to the places where it's needed.

The term "loader" doesn't really mean anything, so we can do whatever we want.

r-b-g-b:
Yeah, after thinking it over I came around to that same conclusion. It's not a "loader," it's a "joiner" which should operate on already loaded things. At least I feel better if I think about it that way :P

r-b-g-b:
Ok a bit of a refactor. I made the loaders ``load_dustrak`` and ``load_gps`` take file contents. ``join`` then takes the already-loaded dataframes, which I think makes a bit more sense.

r-b-g-b:
The funny thing is that now those load functions will be called from the view, and any errors can be raised there. So the title of this PR is now... "Make stuff, like, kinda different"

clintonb (approved):

TangoYankee (approved):

Issue #13, opened 2018-12-20
Add `choices` attribute to `Sensor.unit`


clintonb:
Forcing users to input their units will lead to strange data—spacing, capitalization, misspellings, duplicates. Avoid this issue by defining [`choices`](https://docs.djangoproject.com/en/2.1/ref/models/fields/#choices) for this field.

TangoYankee:
I've been needing this feature in my life!

PR #25, opened 2019-02-06
Eb/choices attribute

Ethan-bradley:
This fixes issue #13, which was to add a dropdown field for the measurements section in the sensor field.


r-b-g-b (approved):
Looks great!

Issue #16, opened 2019-01-27
Use Session as foreign key to Data


r-b-g-b:
A key design consideration for the database was to have each sample in a separate row of the ``Data`` table. An hour session sampled at 1 Hz would have 3600 rows then. 1000 of those sessions would need 3,600,000 rows. To avoid getting too big ``Data`` shouldn't have too many columns. It should also probably not have any many-to-many fields, since they would lead to very large [bridging tables](https://dzone.com/articles/how-to-handle-a-many-to-many-relationship-in-datab). Currently, ``Data`` is very lean: aside from the essential ``lat``, ``lon``, ``time``, and ``value`` fields, it only has a single additional foreign key column to link it to the rest of the database, the ``SessionData`` model. ``SessionData`` contains the raw data file.

But ``Data`` rows can result from two raw data files -- e.g., air quality data coming from the Dustrak csv and the GPS location coming from a separate GPS log. My imperfect solution was to only link each ``Data`` row to the air quality measurement ``SessionData``, conveniently forgetting about the GPS raw data file. Not great. Of course, you could always query other ``SessionData`` from that ``Session`` to find it, but this isn't perfect either, since there could be other unrelated ``SessionData`` entries.

@nbilenko made a suggestion that I hadn't considered, which is to use ``Session`` as the foreign key to ``Data``. Now the database relationship looks like this:

```
Session A --------Data (lat=123.2943, lon=37.1334, value=0.3, session="Session A")
           \------SessionData (upload=dustrak.csv, sensor=<an air quality sensor>, session="Session A") 
            \-----SessionData (upload=gps.log, sensor=<a gps sensor>, session="Session A")
```

This will mean that a separate ``Session`` should exist for each "stream" of data. That is, if a participant walks around collecting two kinds of air quality measurements simultaneously, they will need to make two separate ``Session`` entries, which I think is fine.

clintonb:
What problem are you trying to solve? 3.6M rose is not a lot of data in the grand scheme of things. Data issues will be a nice problem to have, but it’s not a problem that exists today. 

The new schema also makes it impossible to know which data file is linked to a specific datum.

r-b-g-b:
[...]

PR #26, opened 2019-02-06
R b g b/schema rework

r-b-g-b:
Addresses #16
 
I took a somewhat lazy approach to the migrations and just deleted the original migration files and added a brand new ``0001_initial.py``. Since nobody really has any databases up and running yet, this shouldn't cause any problems. If there are objections, I can try to do this maintaining the existing migrations using [this approach](https://docs.djangoproject.com/en/2.1/howto/writing-migrations/#migrations-that-add-unique-fields), but it frankly sounds like a biiiiiig pain.


TangoYankee (changes requested):
From Travis:

`TypeError: 'session_data' is an invalid keyword argument for this function`

TangoYankee (commented):
Two main questions:
1. How are we saving and tracking the Raw GPS files?
2. How do you envision the route fields being used?

r-b-g-b:
1. Raw GPS files are uploaded and saved as ``SessionData`` objects just like the air quality files. They are linked to a ``Session`` and through session to individual data points (``Data``).
2. This came from talks with WOEIP -- there are a few consistent routes that people will walk, and it would be helpful to query on that.

TangoYankee:
>-- there are a few consistent routes that people will walk, and it would be helpful to query on that.

Let's open the handling of the route as a full issue. We proposed it as an idea but never really fleshed out how it would be applied.

For the GPS data, I want to confirm that the 'unit' will be null in the sensor model.

Anyway, I approve of the model changes. I'm mostly clarifying my understanding and doing some sanity checks.

TangoYankee (approved):

r-b-g-b:
I was thinking that he ``Sensor.unit`` field simply indicate that the measurement is a GPS coordinate, e.g., "latlon"

TangoYankee:
Should 'latlon' be one a unit choices, then?

woeip/apps/air_quality/tests/factories.py
```diff
@@ -9,10 +9,20 @@
 from woeip.apps.core.tests.factories import UserFactory
 
 
-class FuzzyLineString(factory.fuzzy.BaseFuzzyAttribute):
+def random_latlon():
```
clintonb: There is no reason to reinvent this. Use `faker`: https://faker.readthedocs.io/en/master/providers/faker.providers.geo.html.

clintonb (commented):
Use Faker for fake coordinates. There is no reason to reinvent the wheel.

woeip/apps/air_quality/tests/factories.py
r-b-g-b: The faker output is a tuple of ``decimal.Decimal``, which isn't compatible with the ``django.contrib.gis.db.models.PointField``. It really really wants the value to be a ``django.contrib.gis.geos.Point``.

r-b-g-b: Specifically...

```
>>> data = models.Data(session=session, value=2.3, time=datetime.datetime.now(), latlon=fake.latlng())
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/django/db/models/base.py", line 470, in __init__
    _setattr(self, field.attname, val)
  File "/usr/local/lib/python3.6/site-packages/django/contrib/gis/db/models/proxy.py", line 76, in __set__
    instance.__class__.__name__, gtype, type(value)))
TypeError: Cannot set Data SpatialProxy (POINT) with value of type: <class 'tuple'>
```

r-b-g-b: I see, you were referring to something else. Addressed below!

woeip/apps/air_quality/tests/test_factories.py
```diff
@@ -0,0 +1,13 @@
+import pytest
```
clintonb: You don't need to test factories. Your tests will do that for you.

woeip/apps/air_quality/tests/factories.py
```diff
@@ -1,18 +1,23 @@
...
 
 from woeip.apps.air_quality.models import Data, Device, Route, Sensor, Session, SessionData
 from woeip.apps.core.tests.factories import UserFactory
 
+fake = faker.Faker()
+
+
+class FuzzyLatLon(factory.fuzzy.BaseFuzzyAttribute):
+    def fuzz(self):
+        return geos.Point(fake.latlng())
```
clintonb: Add to end of line: `# pylint: disable=no-member`

woeip/apps/air_quality/tests/factories.py
```diff
@@ -1,18 +1,23 @@
...
+
 
-class FuzzyLineString(factory.fuzzy.BaseFuzzyAttribute):
+class FuzzyRoute(factory.fuzzy.BaseFuzzyAttribute):
     def fuzz(self):
-        points = [(random.random(), random.random())
-                  for _ in range(20)]
+        points = [fake.latlng() for _ in range(20)]
```
clintonb: Add to end of line: `# pylint: disable=no-member`

clintonb (commented):
Pylint doesn't know about Faker's generators. Those errors are safe to ignore.

woeip/apps/air_quality/tests/test_factories.py
r-b-g-b: Yeah I figured that there was this bug in the code (``DataFactory`` didn't work), so I'd retroactively put a test in there that would have caught that bug. I agree the better solution would be to have tests that used the factories in a more meaningful way, but for now I'd like to keep something in there that would catch any future bugs in the factories.

clintonb: If there is a bug in a factory, it will result in failed tests that use said factory. Retaining this test doesn’t do anything your other tests aren’t already doing.

r-b-g-b: This test does do something our other tests aren't doing, namely test ``DataFactory``. I guess I could write a single test for ``DataFactory`` until we have more comprehensive tests.

clintonb: There is a simpler solution: remove DataFactory. If you don’t have tests that need the factory, you don’t need the factory. Remove it until you need it.

r-b-g-b: Ha, that's a good idea.

r-b-g-b:
Ok, I think all concerns have been addressed. I'll merge this evening unless anyone requests more time to review.

clintonb (approved):

PR #29, opened 2019-02-10
Ty/upload form

TangoYankee:
- Sort imports of `urls`, `settings', forms` and `views`
- Refractor templates, under guidance of [Template naming and structuring conventions](https://oncampus.oberlin.edu/webteam/2012/09/architecture-django-templates)
  * rename `base` as `__base` to reflect its use an extended template
  * rename `home` as `index`, in line with [prevailing convention](https://www.lifewire.com/index-html-page-3466505), move to top level of template directory
  * create `_head` and `_header` as included templates
  * remove error templates, replace with `messages` in header
  * create structure for navigation bar
- Rename `dustrak_upload` as `files` to generalize url and view to
  * several file types (including gps files)
  * several methods (get, post, put, delete)
- In `settings`, set default redirect for login and logout
- In `views`, implement `messages`

Further changes will be made once the database is settled by PR #26


woeip/templates/includes/_header.html
```diff
@@ -0,0 +1,21 @@
+<!-- Navigation Bar -->
+<ul>
+    <li><h1><a href="{% url 'index' %}">West Oakland Air Quality</a></h1></li>
+    <li><a href="{% url 'files' %}">File Management</a></li>
```
r-b-g-b: "File Management" seems like too broad a term -- I'd consider how this sounds to our target audience, the WOEIP users. "Upload a new session" might be more clear.

woeip/apps/air_quality/views.py
```diff
@@ -1,20 +1,22 @@
...
 
-def home(request):
-    return render(request, 'air_quality/home.html')
+def index(request):
+    return render(request, 'index.html')
 
 
-def upload_dustrak(request):
-    """Upload data for a session collected using the Dustrak air quality device and a separate GPS 
+def files(request):
```
r-b-g-b: Similar to below: ``files`` seems to general. ``upload_session`` or ``upload`` is more specific and would be my preference.

r-b-g-b:
Very nice! The template structure is a great addition, as are the flash messages.

woeip/apps/air_quality/views.py
TangoYankee: Thinking about it, a closer replication of standard url naming would be to have the model and then the method. 

So, the view, url, and name would be `upload`, `../files/upload/`, and `upload`

TangoYankee:
The failing tests are related to linting errors of unused imports. These are better resolved before merging into master, rather than merging into upload-form

r-b-g-b:
Agreed, I think we should merge this now and work out the kinks in r-b-g-b/upload-form

r-b-g-b (approved):

PR #38, opened 2019-02-19
Ty/files upload

TangoYankee:
~~Not ready for merge~~
**Ready for Review**

This branch is based on r-b-g-b/upload-form, which adds templates, authentication, views, and urls. for login, home, and file upload. It has been rebased to master. Also, some imports were reordered using isort. Next steps have been outlined with TODO comments which reflect Linting messages and a few suggestions for code changes.


TangoYankee:
@r-b-g-b `views.py` has a coverage score of only 31%. The number will likely change with more commits. However, I think we need tests for views. Should we make tests part of this PR or another one? My vote is making it the next PR. It limits the size of this PR (it's already worth at least two PRs) but avoids accumulating tech debt. 

Possible Reference: [Test-Driven Views and Templates](https://test-driven-django-development.readthedocs.io/en/latest/03-views.html)

r-b-g-b:
Yeah agreed, let's let this PR just be the functionality and submit tests in a separate PR. As long as we all agree that it's wrong and we feel bad for doing it.

r-b-g-b (approved):
This all looks good to me!

woeip/apps/air_quality/dustrak.py
```diff
@@ -115,9 +115,10 @@ def load_dustrak(contents, tz):
     contents = io.StringIO(contents)
 
     header = {}
-    lines = itertools.takewhile(lambda x: x != '\n', contents)
+    lines = itertools.takewhile(lambda x: x != '\r\n', contents)
     for line in lines:
-        line = line.rstrip('\n')
+        print(line)
```
clintonb: `print` is not useful in a production system. If this is important, use `logger.info` to ensure log lines are written to console, files, wherever else logs are routed.

woeip/apps/air_quality/forms.py
```diff
@@ -0,0 +1,17 @@
+from django import forms
+
+from woeip.apps.air_quality import models
+
+
+class SessionForm(forms.ModelForm):
```
clintonb: Given that you are only using `DustrakSessionForm`, I don't see much value in having this base class. Merge it for now, and extract it later if/when you needed.

woeip/apps/air_quality/tests/test_dustrak.py
```diff
@@ -25,12 +25,12 @@ def get_target_data():
 
 def test_joiner():
     """Test the output of the function that joins GPS and air quality measurements based on time stamps"""
-    with open(test_data_directory / 'dustrak.csv', 'r') as f:
```
clintonb, line 28: Why are any changes needed to this file?

woeip/apps/air_quality/views.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def index(request):
+    return render(request, 'index.html')
+
+
+@login_required
+def upload(request):
+    """Upload data for a session collected using the Dustrak air quality device and a separate GPS
+    log file. Creates a new Session instance, two SessionData instances, and Data instances for
```
clintonb: Nit: Consider ditching this second sentence. If your implementation changes, this docstring will probably not be updated, and grow stale. Your first sentence accurately describes what's going on, and is good enough.

woeip/apps/air_quality/views.py
```diff
@@ -0,0 +1,66 @@
...
+    each sample.
+    """
+
+    if request.method == 'POST':
+        form = forms.DustrakSessionForm(request.POST, request.FILES)
+        if form.is_valid():
+            form.save()
+            try:
+                air_sensor = models.Sensor.objects.get(name='Dustrak')
```
clintonb, line 30: What's going on here? Are sensors not specific to individual users? You're going to have more than one Dustrak or GPS sensor.

woeip/apps/air_quality/views.py
```diff
@@ -0,0 +1,66 @@
...
+                                                 sensor=air_sensor,
+                                                 session=form.instance,
+                                                 uploaded_by=request.user)
+                gps = models.SessionData(upload=request.FILES['gps'],
+                                         sensor=gps_sensor,
+                                         session=form.instance,
+                                         uploaded_by=request.user)
+
+                air_quality_contents = force_text(request.FILES['air_quality'].read())
+                _, air_quality_data = dustrak.load_dustrak(air_quality_contents, 'America/Los_Angeles')
```
clintonb: Don't hard code the timezone on the backend. Instead, set it on the frontend.

woeip/apps/air_quality/views.py
```diff
@@ -0,0 +1,66 @@
...
+
+                gps_contents = force_text(request.FILES['gps'].read())
+                gps_data = dustrak.load_gps(gps_contents)
+                joined_data = dustrak.join(air_quality_data, gps_data)
+
+            except (UnboundLocalError, ObjectDoesNotExist, Error) as e:
```
clintonb: What causes these errors? I have never seen `UnboundLocalError`. Do some validation beforehand to catch `ObjectDoesNotExist` and return an error sooner rather than later. `Error` is way too broad. Rather than catch it, let it blow up.

The higher level question I ask myself is, "what is the minimum amount of code that needs to be in this try/except block"?

woeip/apps/air_quality/views.py
```diff
@@ -0,0 +1,66 @@
...
+                messages.add_message(request, messages.ERROR, f'File upload failed, error: {e}')
```
clintonb: Add a logging line—`log.exception()`—so that data is written to the log files. `exception()` has the benefit of including the stack trace, so no need to pass the exception.

woeip/apps/air_quality/views.py
```diff
@@ -0,0 +1,66 @@
...
+                return redirect('upload')
+
+            form.save()
```
clintonb: This is not needed. It's called above.

woeip/apps/air_quality/views.py
```diff
@@ -0,0 +1,66 @@
...
+            dustrak.save(joined_data, form.instance)
+
+            messages.add_message(request, messages.SUCCESS, 'Files successfully uploaded')
+            return redirect('upload')
+
+    else:
+        form = forms.DustrakSessionForm(initial={'collected_by': request.user})
+
+    return render(request, 'upload.html',
+                  {'user': request.user, 'form': form})
```
clintonb: One line. This doesn't need to be broken up.

woeip/apps/air_quality/views.py
```diff
@@ -0,0 +1,66 @@
...
+                joined_data = dustrak.join(air_quality_data, gps_data)
+
+            except (UnboundLocalError, ObjectDoesNotExist, Error) as e:
+                messages.add_message(request, messages.ERROR, f'File upload failed, error: {e}')
+                return redirect('upload')
+
+            form.save()
+            air_quality.save()
+            gps.save()
+            dustrak.save(joined_data, form.instance)
```
clintonb: Extract `form.instance` to a variable. It's used in three different places.

woeip/apps/air_quality/views.py
```diff
@@ -0,0 +1,66 @@
...
+
+            messages.add_message(request, messages.SUCCESS, 'Files successfully uploaded')
+            return redirect('upload')
+
+    else:
+        form = forms.DustrakSessionForm(initial={'collected_by': request.user})
```
clintonb: Extract `request.user` to a variable. It is used in four different places.

woeip/templates/includes/_head.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html lang="en">
```
clintonb: Although this project probably won't be translated into other languages and locales, it's still a good practice to maintain good i18n/l10n habits:

```django
{% load i18n %}
{% get_current_language as LANGUAGE_CODE %}
<!DOCTYPE html>
<html lang="{{ LANGUAGE_CODE }}">
```

woeip/templates/includes/_header.html
```diff
@@ -0,0 +1,21 @@
...
+    <li><h1><a href="{% url 'index' %}">West Oakland Air Quality</a></h1></li>
+    <li><a href="{% url 'upload' %}">File Upload</a></li>
+    <li>
+        {% if user.is_authenticated %}
+            {{user}}: <a href="{% url 'logout' %}">Logout</a>
+        {% else %}
+            Guest: <a href="{% url 'login' %}">Login</a>
+        {% endif %}
+    </li>
+    <li><a href="mailto:contact_email">Request Account</a></li>
```
clintonb, line 12: Should this be a valid email address?

woeip/templates/includes/_header.html
```diff
@@ -0,0 +1,21 @@
+<!-- Navigation Bar -->
+<ul>
+    <li><h1><a href="{% url 'index' %}">West Oakland Air Quality</a></h1></li>
+    <li><a href="{% url 'upload' %}">File Upload</a></li>
+    <li>
+        {% if user.is_authenticated %}
+            {{user}}: <a href="{% url 'logout' %}">Logout</a>
+        {% else %}
+            Guest: <a href="{% url 'login' %}">Login</a>
```
clintonb: There are no guest accounts, so remove `Guest:`.

woeip/templates/__base.html
```diff
@@ -0,0 +1,7 @@
+{% include 'includes/_head.html' %}
```
clintonb, line 1: The file name only needs one underscore: `_base.html`.

woeip/templates/upload.html
```diff
@@ -0,0 +1,8 @@
+{% extends '__base.html' %}
```
clintonb: Nit: Move this to `woeip/apps/air_quality/templates/air_quality/upload.html`. This keeps your templates organized so that they are easy to manage as you introduce more templates and apps.

woeip/urls.py
```diff
@@ -1,9 +1,13 @@
...
+from django.urls import include, path
 
+from .apps.air_quality.views import index, upload
 from .apps.core.views import health
 
 urlpatterns = [
+    path('', index, name='index'),
     path('health/', health, name='health'),
     path('admin/', admin.site.urls),
+    path('accounts/', include('django.contrib.auth.urls')),
```
clintonb, line 9: Nit: keep this list alphabetized.

woeip/urls.py
```diff
@@ -1,9 +1,13 @@
...
+    path('files/upload/', upload, name='upload')
```
clintonb: Nit: add a trailing comma

woeip/templates/upload.html
TangoYankee: I'd like to keep all the templates centralized in `woeip/templates`. The templates are dependent on each other's structure. It makes sense to keep them together, to make their dependencies easier to navigate. I created a `woeip/templates/air_quality` folder instead. This can be revisited, if this organization style becomes cumbersome.

woeip/apps/air_quality/tests/test_dustrak.py
TangoYankee: @r-b-g-b Would be better to speak to that.

woeip/apps/air_quality/views.py
TangoYankee: This is actually a topic of discussion among the WOIEP team. You can see it at PR #40 and Issue #43 . I think it belongs in its own PR. In my opinion, this current PR is too large and unfocused already. I'd like to keep this sensor code for now, simply to get the agreed upon changes into the codebase.

woeip/apps/air_quality/views.py
TangoYankee: The errors are actually related. `UnboundLocalError` is caused when the `air_sensor` and `gps` variables are called without being assigned a value. They fail to be assigned when there is an `ObjectDoesNotExist`. I've struggled to find guidance on good 'try/except' methodology. This comment helps a lot.

woeip/templates/__base.html
TangoYankee: I've seen and used the style of one underscore in the past. For this project, I looked for established naming guidelines and couldn't find them. I stumbled on a [blog from 2012](https://oncampus.oberlin.edu/webteam/2012/09/architecture-django-templates) that recommended naming *includable* templates with a single underscore and *extendable* templates with a double underscore. I'd like to try this naming convention, unless they're outdated and other guidelines have become standardized.

woeip/templates/includes/_header.html
TangoYankee: Soon. I'll open an issue to make sure we come back to it, once the WOEIP team has discussed it.

Edit: Issue #47

TangoYankee:
I'm currently short on time. I'm making a commit to address the changes I can resolve now. I still need to address: 
- ~~Time Zone~~
- ~~Try/Except~~
- ~~Logging~~
- ~~i18n/l10n~~

Edit: Found some time

woeip/apps/air_quality/views.py
TangoYankee: From the `logging` documentation, project code, and error messages, it seems the exact syntax should be `logger.exception(e)` or `logging.exception(e)`. Unfortunately, I was unable to confirm this because I also could not find the log file. From reading `settings.py`, shouldn't it be towards the base directory of the project?

Edit: From a previous project I helped with ([openoakland/trashtalk](https://github.com/openoakland/TrashTalk/blob/master/trashtalk/apps/cleanups/views/template_views.py)), the answer is `logger.exception('message to user')`. However, I still can't find the logging file.

woeip/apps/air_quality/views.py
TangoYankee: After thinking it over, my understanding for exception handling is:
1) Understand what can go wrong
2) Code defensively, to keep it from going wrong
3) Add exception handling to catch known errors that are difficult to prevent/out of the developer's control

As it applies to this case, the only known error is related to loading the dustrak and gps sensor information. Because the process is not yet understood, it's impractacle to code defensively/add validation. Consequently, I modified the code to catch the known error and let the rest of the view 'blow up'. Validation will be added in a later PR, when the dustrak/gps upload process is established.

woeip/apps/air_quality/views.py
clintonb: Logs are written to the `logs` directory at the repo root, assuming `ENABLE_LOGGING_TO_FILE` is `True`. Otherwise, logs are only written to the console.

https://github.com/openoakland/woeip/blob/master/woeip/settings.py#L187

TangoYankee: It seems my local system objected to Django creating a file and writing to it, while woeip was running in a Docker Container. I changed [Line 202](https://github.com/openoakland/woeip/blob/master/woeip/settings.py#L202) to `/app/woeip/app.log`, to follow a path used by the container.

I imagine that someone running woeip directly on a local system (not using Docker) would not encounter this issue. For them, instead of fixing an issue, my addition of `/app/woeip/` would bury the log file within two folders.

clintonb: Exactly how did your local system object? What did you set? What was the resulting output?

The code should work in all cases. If that is not true, it needs to be fixed.

TangoYankee: * Within `docker-compose.local.yml`, under `app` -> `environment`, I created the variable `ENABLE_LOGGING_TO_FILE` and set it to `True`
* I haven't been receiving errors in the terminal and can't provide one in this case, either.
* The expected behavior was that an app.log file would be created and populated with a known error
* The observed behavior was that no such file was created
* I modified `settings.py`, as described above. This created the expected behavior.

Would it be better to modify line 179 of `settings.py`... `'filename': environ.Path(LOG_FILE_PATH).path(filename)` ?

Ethan-bradley:
Is numpy (used in dustrak.py) supposed to imported in the pipfile?

TangoYankee:
@Ethan-bradley Why wouldn't it be?

Ethan-bradley:
@TangoYankee When I downloaded this PR on my computer, numpy seemed to be missing from the pipfile. However, it still seemed to work on Jess's computer without having to add numpy to the pipfile.

TangoYankee:
`Numpy` is in `pipfile.lock`- it makes sense that the code still works. However, it's strange that the `pipfile.lock` contains `numpy` in the first place, when neither `master` nor this branch contain `numpy` in their `pipfile`. Could you open an Issue about `numpy "~=1.16.1"` missing from the `pipfile` @Ethan-bradley ?

Ethan-bradley:
Sounds good. I think it could be added into pipfile the next time there is a major commit to this branch.

clintonb:
Lurking: just fix it. No need for an issue or waiting. Just update the Pipfile and lock file. It’s obviously a bug on master. It’s easy to fix, rather than risk someone else encounter the bug and waste time trying to fix it.

Ethan-bradley:
Sure

TangoYankee:
@clintonb Wilco. My concern was that it's a symptom of a bigger problem. But, it seems I'm being too conservative on this one. It's either isolated or there is no need to worry about a bigger problem- unless it starts creating other issues.

woeip/apps/air_quality/tests/test_dustrak.py
TangoYankee: These changes are part of a [PR #39](https://github.com/openoakland/woeip/pull/39#issue-254468163), which Robert made to this branch. As far as I can tell, the file isn't stored in binary and there is no reason to open it in binary for the test.

Edit: After attempting to run the previous version of the test, the change to binary is definitely necessary.

PR #39, opened 2019-02-20
Use current versions of dustrak and gps loaders in upload view

r-b-g-b:
Prior to this change, file upload does not work.
 - Detect ``\r\n`` line endings in the raw data txt files
 - Dustrak and gps loaders want strings, not file handles
 - ``dustrak.save`` wants a ``Session`` instance, not ``SessionData`` reflecting changes since #26 

I fixed this in some other branch, but somehow it got lost, so reapplying here.


TangoYankee (approved):
* Travis Checks are failing on linting issues. These issues will be resolved before merging into master.
* Functional tests are passing.

Issue #4, opened 2018-12-08
Docker Curriculum


TangoYankee:
An intro for the confused: [Docker Curriculum](https://docker-curriculum.com/)

Issue #2, opened 2018-11-29
Readme Description


TangoYankee:
@clintonb In reference to the readme TODO, a (not brief) description exists on a branch of my fork for the previous woeip repo. [Draft Readme](https://github.com/tangoyankee/woeip/tree/readme-contributing)

PR #1, opened 2018-11-28
Cleanup

clintonb:
- Updated README badge URLs
- Updated Docker image name
- Updated Dockerfile
- Using environment variable to instruct pipenv to use system Python.
- Updated Terraform backend bucket name

PR #40, opened 2019-02-20
Infer Dustrak sensor from header

r-b-g-b:
Instead of hard-coding the Dustrak sensor in the upload page, infer it from the header. Since could involve the creation of several new database entries, wrap the database-interacting code in a django [transaction](https://docs.djangoproject.com/en/2.1/topics/db/transactions/#tying-transactions-to-http-requests), so that if an error occurs during upload, no changes to the database are made.

### DIFF
--- a/woeip/apps/air_quality/views.py
+++ b/woeip/apps/air_quality/views.py
@@ -1,15 +1,13 @@
 # TODO: imports pylint lists as unused: datetime, authenticate, login, User
-import datetime
 import logging
 
 from django.contrib import messages
-from django.contrib.auth import authenticate, login
 from django.contrib.auth.decorators import login_required
+from django.db import transaction
 from django.shortcuts import redirect, render
 from django.utils.encoding import force_text
 
 from woeip.apps.air_quality import dustrak, forms, models
-from woeip.apps.core.models import User
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


woeip/apps/air_quality/views.py
```diff
@@ -24,49 +22,44 @@ def upload(request):
     log file. Creates a new Session instance, two SessionData instances, and Data instances for
     each sample.
     """
-
+    tz = 'America/Los_Angeles'
```
r-b-g-b, line 25: Hard-coded for now. Will need to add a form element at some point.

TangoYankee (changes requested):
This code seems to be based on these TODOs:
`forms.py`
>  TODO: Include Sensor fields that are seen by client when uploading the files.
> 1) It is more transparent to the user what data is being loaded and why.
> 2) It is easier to generalize the website, once other sensors are available.

`views.py`
> TODO: Rather than hardcode the value of the Sensors, present it as an option in
> the form the client sees. However, have it pre-filled with the pre-loaded objects
> of Dustrak and GPS. This will also achieve the goal of outsourcing some of the
> view code to other modules. ie) The object call will be done in the form, instead

From this PR, I'm inferring that I wrote these TODOs without a full understanding of your intention with the dustrak and gps values. 

My intent was that the user experience would include a dropdown on the client side, where they could select which sensors they used. However, your intention seems to be to automatically detect the type of sensor that is being used. This also includes values which will change, such as the calibration date. 

This is greatly increasing the scope of the ty/files-upload PR. It's also getting into questions of UX/UI. Rather than tackle this problem now, let's create an issue to discuss the direction of the code. @theecrit, we need you in discussion, to get the UX/UI perspective. 

For ty/files-upload, we'll leave the current solution in place. I'll remove the TODOs from the code and instead place them in the issue we're opening to address this PR. This will reduce the scope of the ty/files-upload PR. Let's focus on getting ty/files-upload to pass the rest of the tests.

theecrit:
Yeah, I need to get this installed locally so I can fully understand the current state, and then consider that against the potential options y'all are discussing. I agree that we should keep scope as lean as possible at this stage, provided it doesn't overly impact usability.

We currently have only one device model to accommodate, but there is a possibility of future models. So then it becomes a matter of:

1. _Is it a better user experience_ to have the system identify the device instead of the user?
2. If so, is it _technically feasible_ for the system to identify device model without user input?
3. If so, what is the _level of technical effort_ to do so?
4. Is that effort _worth the payoff_ in greater usability?

I believe the answer to #1 is yes. 

I suspect the answer to #2 is "possibly, if the devices output some sort of unique data point that's system-readable." This requires us to know what devices are likely to be used, and what those data points look like. We could discuss this with WOEIP to understand the likelihood of additional device models.

I suspect the answer to #3 is "moderate," but that would depend on #2, and y'all are better equipped to answer that.

I suspect that #4 is "yes," _once a second model is introduced in use._ Until then, can we build the system so it assumes the current model is in use, and then backlog the additional requirement to self-detect additional models? That backlog card would actually be two items: 1) research different device's data output, and then 2) identify technical feasibility of auto-detection.

Long winded. Sorry, but I think this is a good example of how we can make these decisions more objectively.

r-b-g-b:
This PR accomplishes for 3 the Dustrak sensor. It's relatively straightforward: the dustrak log header contains all of the information we need to identify which sensor was used (partially because we designed the database tables by inspecting the header, but oh well...). This is especially simple because the header is machine generated and will never need any fancy parsing.

Other devices will need more or less manual input depending on what is stored in their headers. We can discuss how to design the user interactions when we handle multiple sensors, but I think it will be within reach in terms of technical effort.

TangoYankee:
Automating the upload process makes sense to me, as long as we communicate the process to the user. Though, let's move this discussion to Issue #43. 

My intention is to discuss problems and solutions in GitHub Issues or Trello Cards. That way, we know what the code should look like, before writing it. Then, discussions in PRs can be limited to quick suggestions on how to improve the implementation of agreed upon solutions.

theecrit:
Closing the PR, but let's keep the branch to maintain that work.

Issue #76, opened 2019-03-27
As a user, I want my starting point to be the sign-in page


theecrit:
**Is your feature request related to a problem? Please describe.**
The user needs a starting URL to begin their experience.

**Describe the solution you'd like**
Redirect /index to /signin (I'd prefer to use "sign in" language consistently over "log in" because it's more user-centric, as opposed to system-centric). 

**Describe alternatives you've considered**
The resources to scope, design, and implement a proper home page at /index is better used for core MVP functionality. 

**Additional context**
We've got a [backlogged issue](https://github.com/openoakland/woeip/issues/34) for the /index landing page once MVP has shipped. This will also buy us time to finalize some of the open questions around project positioning and team roles.

**Edit:** `Sign in` already redirects to `index`. If we have `index` redirect to `signin`, then it will be an endless circle. Alternatively, we could redirect from `index` to `upload`. If the user is logged in, they continue seemlessly to `add data`. If they are not signed in, they are gracefully redirected to the sign in page. Once signed in, they are advanced to the `upload` page. This is bad django. But, it's good UX. We'll pay for our sins, after MVP. Changes are updated in the **Technical Resources** Section
-Tim

### For Developers
**Affected Components**
`air_quality/views.py` `index` function

**Technical Resources**
- The index view should `return redirect ('upload')` ~~`redirect('[insert name of sign in page]')`~~
- Alternative approach from simple is better than complex [Django Login Views](https://simpleisbetterthancomplex.com/tutorial/2016/06/27/how-to-use-djangos-built-in-login-system.html)
- [StackOverFlow](https://stackoverflow.com/questions/20208562/homepage-login-form-django#20210318)
**Design Documentation**
TBD

TangoYankee:
I swapped the tags. If I'm not mistaken, the landing page is **backlog** but this redirect is **MVP**. Also, I think it's ready from implementation.

theecrit:
@TangoYankee When you say we have /signin redirecting to /index, do you mean once the user is signed in, they are redirected to /index?

TangoYankee:
> @TangoYankee When you say we have /signin redirecting to /index, do you mean once the user is signed in, they are redirected to /index?

Currently, yes. However, if they originally asked to go to `add data` but are stopped by the `signin` page, then signing in will redirect them to `add data`. That's the behavior I'm going for, by have `index` redirect to `add data`. If I'm not mistaken, that's the original flow you designed.

TangoYankee:
[...]

PR #85, opened 2019-04-11
redirect index url to upload page

TangoYankee:
## Checklist
- [x] Add description
- [x] Pass linting check
- [x] Pass functional tests
- [x] Request code review
- [x] Merge own code

## Description
Issue: #76

Redirect the index url to the upload page. Unauthorized users will have a chance to sign in, before being redirected to the upload page.

### DIFF
--- a/woeip/apps/air_quality/views.py
+++ b/woeip/apps/air_quality/views.py
@@ -12,7 +12,10 @@ logger = logging.getLogger(__name__)
 
 
 def index(request):
-    return render(request, 'index.html')
+    """Logged in users are directed to the upload page. Signed out users are directed to the login,
+    which then directs to the upload page
+    """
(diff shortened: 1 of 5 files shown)


nbilenko:
I would suggest using `''` as the proxy for `upload` for now, since we expect the users to come to the upload page when they come to the site (which can be changed later if we expect an index page post-MVP). That would avoid a redirect from one view to another view.

TangoYankee:
> I would suggest using `''` as the proxy for `upload` for now, since we expect the users to come to the upload page when they come to the site (which can be changed later if we expect an index page post-MVP). That would avoid a redirect from one view to another view.

Wilco

nbilenko (approved):
LGTM!

Issue #71, opened 2019-03-20
As a user, I want to be able to sign into my account.


theecrit:
**This issue covers the following features/flows**
- Account sign-in user interface (html template and CSS/JS styling)
  - Backend is already handled by Django built-in functionality
- Account creation
  - "Account creation" should be changed to read "contact [insert email address] to request an account"
  - It should be plain text, rather than a hyperlink 
- Password hide/reveal
  - See note below

**Not in this issue's scope***
- Password reset / `Forgot Password` link
- Footer
- `Visit WOEIP` link

**Notes on "Sign In" button**
- Ignore the "inactive state" in the wireframes.

**Notes on password hide/reveal**
- Password field on sign-in form to default to obscuring the characters as they're entered.
- Visual indicator/toggle to reveal characters (either eye icon right-aligned within input field, or a text label reading "Show", that changes to "Hide" on reveal).

**Additional context**
~n/a~
Edit: This is an enhancement to the built-in Django Form. - Tim

### For Developers
**Affected Components**
The login template and any CSS and/or JS needed to style it

**Technical Resources**
[Django Forms](https://docs.djangoproject.com/en/2.1/topics/forms/) will need to be modified with css classes from Foundation
~~[Zurb Foundation Show Password](https://foundation.zurb.com/building-blocks/blocks/show-password.html)~~
[W3 Schools Javascript Simple Reveal](https://www.w3schools.com/howto/howto_js_toggle_password.asp)
- [Forms Example](https://docs.djangoproject.com/en/dev/topics/auth/default/#module-django.contrib.auth.views)

**Design Documentation**
[WOAQ-Wires-PasswordReveal-20180326.pdf](https://github.com/openoakland/woeip/files/3011419/WOAQ-Wires-PasswordReveal-20180326.pdf)

theecrit:
Interesting [article from Luke Wroblewski](https://www.lukew.com/ff/entry.asp?1941/) on defaulting to show password, with [backup from Nielsen Norman](https://www.nngroup.com/articles/stop-password-masking/). I'm not married to either approach, but seems that defaulting to show password, with an option to hide, is more low-computer-skills friendly. And since security is not a primary part of the UX here, my inclination would be to default to making it easier for the user.

[...]

PR #87, opened 2019-04-14
Signin page styling

TangoYankee:
## Checklist
- [x] Add description
- [x] Pass linting check
- [x] Pass functional tests
- [x] Request code review
- [ ] Merge own code

## Description
Issue: #71 

Responsive sizing of signin form, text to give users context, password toggle, email to request account.


theecrit:
See my comment on #71, please and thanks!

nbilenko (changes requested):
For the h1/button/general element styling, does that need to be localized? It seems like the style can be general across the site.

TangoYankee:
> For the h1/button/general element styling, does that need to be localized? It seems like the style can be general across the site.

@theecrit This is a question of implementing your design. Do you want the links in the navigation bar to have their style extended to all buttons in the site?

theecrit:
Hi all, sorry for the delay. I'm working on figuring out how to get the latest build on my machine so I can see where we're at, but to try and answer the questions above:

> For the h1/button/general element styling, does that need to be localized? It seems like the style can be general across the site.

Yes, we probably want it generalized into a site-wide style guide. Of course, we don't have one yet but I created a [Trello card for it](https://trello.com/c/fGvq6ou6). For now, what specifics do you need from me to unblock this? Type styles and button styles?


> Do you want the links in the navigation bar to have their style extended to all buttons in the site?

Probably not. Navigation menu should be styled differently than action buttons. Of course, nav links should be styled consistently across the site (i.e. from page to page).

theecrit:
For now, how about we use the button styles in the latest incarnation of https://cfa-styleguide.herokuapp.com/cfa/styleguide (provided by Tom as a work-in-progress for brigade projects)?

TangoYankee:
Upon rereading Natalia's comment, I think I misinterpreted the part to which see was referring. 

The question really seems to be whether to apply styles only to the current page, or extend them across the site to all elements with those tags. @nbilenko I want to default to localizing styling elements. This reduces the chance of unintentionally affecting another component.

For the CfA button styling guide, I'll implement the wireframe styling to the best of my ability. Below is the current iteration. On hover, the Signin form button turns black with white text. The right side links of the nav bar turn green. @theecrit, If you'd like to change them to the CfA styling, update the wireframes and I'll update the design.

WOEIP Signin

theecrit:
A couple of things:

I’d really advise against localizing styling. Style guided and design systems exist for a reason, and trying to update every single instance of a given style should a change be needed is not tenable. Let’s come up with a better solution. Unless I’m misunderstanding the questions, as well. 

Re: wireframe updates. I’d like us to practice uncoupling wireframes from redlines (pixel-specific visial design). Again, it’s about finding the path of least resistance balanced with best practice: the visual design styles should remain independent of content. The wireframes are intended to communicate content, hierarchy, relative layout, and in some cases, behavior. Visual design styles (e.g. css) need not be (and shouldn’t be) tied to wires, or we’ll add an additional layer of documentation labor that will be prohibitive to shipping. That’s why design systems like the CfA style guide exist; so developers can apply consistent styling without needing redlines. I’m happy to specify which styles should applied to specific elements in the wires, but I’m not going to be able to skin the site as we go. Of course, maybe that’s what you meant in the first place? 

Thanks!
J

TangoYankee:
It sounds like we're identifying a few differing assumptions. They're leading to a few things we need to discuss, that are outside the scope of a Pull Request discussion.

1) Wireframing v Redlining
2) Style guidelines (Default Foundation or CfA)
3) Scoping preferences for custom styling
2) The placement of buttons in the navigation bar

TangoYankee:
See issue #92 for explanation and discussion of changes

theecrit:
The main outstanding issue I'm seeing right now (if I understand the implementation correctly) is that buttons as navigation are problematic for a number of reasons. The main one is that buttons should be used as a _control_ interaction, whereas navigation is about wayfinding via _links_. Best practice, as far as I understand it, is to use unordered lists and CSS styling for navigation menus. Examples:

- https://a11y-style-guide.com/style-guide/section-navigation.html#kssref-navigation-navigation-basic
- https://www.w3.org/WAI/tutorials/menus/structure/

That A11y Style Guide link also includes a mobile guideline, which stipulates a `<button>` element for the hamburger icon (presumably because it's a _reveal_ control). I'd imagine it would then reveal a `<ul>` based menu.

r-b-g-b (approved):
Looks good to me!

woeip/templates/registration/login.html
```diff
@@ -1,8 +1,49 @@
...
+    of West Oaklanders</p>
+    </div>
+  </div>
+    <div class="grid-x">
+      <div class="cell medium-6"><h3>Sign in to upload data</h3></div>
+    </div>
+    <div class="grid-x">
+      <div class="cell medium-6">
+        <p>Or request an account from the WOAQ team at 
+          <a href="mailto:woaq@openoakland.org">woaq@openoakland.org</a>
```
nbilenko, line 21: The email should probably be passed from the backend as a context variable, but that can happen in the future.

nbilenko (approved):

Issue #81, opened 2019-04-01
As a user, I want to upload air quality and location data to the website


TangoYankee:
**Is your feature request related to a problem? Please describe.**
After collecting air quality data, I would like to upload it to WOAQ to be stored. 

**Describe the solution you'd like**
I will have two files:
1) A csv file which contains air quality measurements taken by a dustrak sensor
2) A log file which contains GPS location data taken by a ....gps?... @theecrit, what's the device?

As a signed in user, I can navigate to an `upload`/`add data` page. On this page, there is a button `find files on my computer`. Clicking this button will open a file explorer where I can select the correct files. I will receive feedback on whether the process was a success. If it was a failure, I will be told and given another chance.

**Describe alternatives you've considered**
Previous version had numerous fields on the form, each for a different value. However, having the user input each value is considered too tedious. It may even be overwhelming to ask so many questions. This method is preferred because the data will be parsed automatically.

**Additional context**
This issue is a prerequisite for issue #31. It provides the backed structure to the upload process.
In reference to the wireframes, it does not achieve the *drag and drop* functionality. However, it should achieve the *find files on my computer* functionality. Issue #63 depends on this process. It will be passed the URIs of the files in the S3 bucket.

**Labels**
- [x] Add a status label
  - `backlog`
  - `discussing`
  - `ready`
  - `implementing`
- [x] Add additional labels as needed

### For Developers
**Affected Components**
- `form` for upload
- `view` for upload
- `template` for upload

**Technical Resources**
[Django Forms](https://docs.djangoproject.com/en/2.1/topics/forms/)
[Django Messages Framework](https://docs.djangoproject.com/en/2.1/ref/contrib/messages/)
[Multiple File Upload tutorial](https://simpleisbetterthancomplex.com/tutorial/2016/11/22/django-multiple-file-upload-using-ajax.html)
[Multiple FIles Tutorial Source Code](https://github.com/sibtc/multiple-file-upload)
[Django Files Tutorial](https://simpleisbetterthancomplex.com/tutorial/2016/08/01/how-to-upload-files-with-django.html)

**Design Documentation**
[Wireframes V3](https://github.com/openoakland/woeip/files/2959738/WOEIP-UploadFlow-v3.pdf)

- Input for this issue: the two user-uploaded files
- Data processing in this issue: store the data files in the S3 bucket
- Output for this issue:
[...]

PR #109, opened 2019-06-13
Upload files to database, no drag and drop

TangoYankee:
## Checklist
- [x] Add description
- [x] Pass linting check
- [x] Pass functional tests
- [x] Request code review
- [x] Merge own code

## Description 
Issue: #81 

### Significant Changes
- Added GPS File field to `SessionData` Model/Table
  - This field is required. It has a default value of a blank string. 
  - The default value is required to facilitate a database migration
  - By adding a field for GPS files, the CSV and Log files can be paired together under a single unique ID in a persistent location that is easily accessible (the database)
  - Making it required enforces the upload form on the server side
  - **Next steps:** enforce which file goes into which field
- Made the relationship between `session` and `sessionData` _optional_.
  - Keeping the relationship mandatory would've required changing the workflow. 
    - `Session`s would need to be created before creating `SessionData`s
    - The GPS and DusTrak files are kept in `SessionData`
  - The relationship still exists and can be set after the data is uploaded
  - The `session.date_collected` field was removed from the `name` of `sessionData`
    - There is no longer a guarantee that date_collected _exists_
  - **Possible Future Concern**
    - The `Data` table/model relies on searching by `session`
    - Need to be careful that the `session` object exists and has the correct relationships, before populating the `data` table
- Refactored `Data Form`
  - Changed form name for style preference
  - Removal of extra fields to focus solely on the upload of files
  - Added of Zurb Foundation classes to input field.
    - This makes a simple drag-and-drop capability. 
     - It is not meant to replace the full drag and drop functionality
     - It's a quick stand-in (that I honestly found accidentally)
- Rewrote most of the `Views`
  - Changed to Class-based views 
    - They have more built-in functionality than function based views
   - Removed all processing of the data. 
     - The work lives on in other branches and will likely be used in future issues.
     - However, it is not currently contributing. It was removed to focus on the functionality of file upload
- Added `ViewSessionData` page
  - It is a temporary development-only page
  - It provides a page where developers can get feedback on the state of the `SessionData` table
  - All support code relating to the view, page, links, and url are meant to be changed or removed on the path to MVP

### Issue
- Attempting to upload a 2MB file results in failure
  - This needs to be turned into a GitHub Issue/Bug Report
- Linter needs to be adjusted
  - added `no-else-return` to list of disabled warnings


Ethan-bradley:
For the 2MB size file problem, is that if you upload a file larger than 2MB or if the file size just happens to be 2MB?

woeip/urls.py
```diff
@@ -1,12 +1,13 @@
...
+from .apps.air_quality import views
 from .apps.core.views import health
 
 urlpatterns = [
-    path('', upload, name='upload'),
+    path('', views.Upload.as_view(), name='upload'),
     path('accounts/', include('django.contrib.auth.urls')),
     path('admin/', admin.site.urls),
     path('health/', health, name='health'),
+    path('view_session_data', views.ViewSessionData.as_view(), name='view_session_data'),
```
nbilenko: Suggest naming this something like `view` for url simplicity.

woeip/apps/air_quality/views.py
```diff
@@ -1,61 +1,40 @@
...
+        else:
+            messages.add_message(self.request, messages.ERROR, 'File upload error')
+            path = redirect('upload')
+        return path
+
+
+class ViewSessionData(View):
+    """Provide temporary development page to view all uploaded SessionDatas."""
+    def get(self, request):
+        sessionData_list = SessionData.objects.all()
```
nbilenko: please use a consistent case (snake case for python variables), in this instance `session_data_list`.

woeip/apps/air_quality/views.py
```diff
@@ -1,61 +1,40 @@
...
+    def get(self, request):
+        """Present file upload form to user"""
+        return render(self.request, 'air_quality/upload.html', {
+            'form': SessionDataForm
+        })
+
+    def post(self, request):
+        """Save files to SessionData table"""
+        files = self.request.FILES
+        form = SessionDataForm(self.request.POST, files)
```
nbilenko, line 24: A more standard syntax for this section would be (esp. better to avoid the circular redirect)
```
    if form.is_valid():
        form.save()
        return redirect('view_session_data')
    else:
        messages.add_message(self.request, messages.ERROR, 'File upload error')
return return render(self.request, 'air_quality/upload.html', {
    'form': SessionDataForm
})

File: woeip/apps/air_quality/migrations/0004_auto_20190613_0043.py [hunk 4]
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.7 on 2019-06-13 00:43
```
nbilenko, line 1: Could the migrations be compressed into one (by running migrate once on a fresh master version of the database)? They are a bit redundant.

woeip/apps/air_quality/models.py
```diff
@@ -54,19 +54,20 @@ def __str__(self):
 
 
 class SessionData(TimeStampedModel):
-    """The raw data file generated during a session. Assumes one and only one file per sensor,
-    although multiple sensors can be linked to one session"""
+    """The raw data files generated during a session. Assumes one general and one gps file.
+    Multiple sensors can be linked to one session"""
     upload = models.FileField(upload_to='session_data')
+    upload_gps = models.FileField(upload_to='session_gps_data', default="")
```
nbilenko: Suggestion for improving variable naming clarity: `sensor_file` and `gps_file` instead of `upload` and `upload_gps`.

woeip/apps/air_quality/forms.py
```diff
@@ -1,14 +1,21 @@
...
-        model = models.Session
-        fields = ('collected_by', 'date_collected', 'route', 'air_quality', 'gps', 'timezone')
+        model = models.SessionData
+        fields = ('upload', 'upload_gps')
+
+    def __init__(self, *args, **kwargs):
+        """Zurb Foundation classes expand the area of the input fields,
+        facilitating simple drag-and-drop functionality"""
+        super(SessionDataForm, self).__init__(*args, **kwargs)
+        self.fields['upload'].widget.attrs.update({
```
nbilenko: Suggestion:
```
for field in ['upload, 'upload_gps']:
    self.fields[field].widget.attrs.update({
        'class': 'callout primary large text-center'
    })
```

nbilenko (commented):
These changes look good. I have a few suggestions. Also, I wasn't able to check out the branch to test - could you push to remote if you haven't yet? Thank you!

TangoYankee:
> These changes look good. I have a few suggestions. Also, I wasn't able to check out the branch to test - could you push to remote if you haven't yet? Thank you!

Hmm, I double checked that it's in the list of openoakland/woeip list of branches. I'm not sure why it isn't letting you check it out

nbilenko:
My bad, I forgot to pull. Will check it out shortly.

TangoYankee:
> For the 2MB size file problem, is that if you upload a file larger than 2MB or if the file size just happens to be 2MB?

Good question. I haven't explored the issue beyond the one error that I received. In that case, the file was listed as 2MB. This question should be part of the issue/bug report, when it's opened.

woeip/urls.py
TangoYankee: wilco

woeip/apps/air_quality/views.py
TangoYankee: Yeah... I figured it was too weird. I'll be normal and change it.

woeip/apps/air_quality/migrations/0004_auto_20190613_0043.py
TangoYankee: I agree. Though, I didn't want to invest too much time addressing it; I don't think it significantly hinders performance. 

That said, if you know a quick/straightforward method, then let's do it!

woeip/apps/air_quality/models.py
TangoYankee: I'll add it to the list of redundant migrations :stuck_out_tongue:

woeip/apps/air_quality/forms.py
TangoYankee: Slick! I'll give it a shot

woeip/apps/air_quality/views.py
TangoYankee: - I like your suggestion to render the upload page again, rather than redirecting. I'll incorporate that change.
- As far listing return twice, that's how I originally structured it. Unfortunately, it causes a linting error. This raises a dilemma between suppressing linting warnings or going against django convention. I ended choosing the later for two reasons. 
 1. I think suppressing warnings is a dangerous game
 2. My freshman python instructor taught me that functions should only have one return. Django be damned, I'm  following my instructor!

Anyway, I'd like to keep the 'path' style with a single return. But, if you're adamant about two returns, then I'll change it.

nbilenko: I've heard of single return style for other languages but it isn't commonly used for Python - neither PEP8 nor Google style guide prescribe it. Since Python does automatic garbage collection, there isn't a programmatic downside to doing multiple returns, except in cases where it makes the code easier to read not to do them. In this case, I think having multiple returns makes the code easier to read since as you mentioned, multiple returns are idiomatic to django (e.g. in the [docs for class-based views](https://docs.djangoproject.com/en/2.2/topics/class-based-views/intro/#handling-forms-with-class-based-views)), and it's usual syntax in django views to return the name of the view in the return statement.

One of the linters we're using (pylint) is currently set to the default settings (in the `pylintrc` file). This sometimes doesn't work well for a specific use case like web development (e.g. another example is that there is a rule against too many ancestors, which doesn't work well for django, which nests a lot of imports). It's pretty common to configure specific linter settings per codebase, and I think it's quite reasonable for us to modify these settings if it's sensible.

Btw I realized I had an extraneous `else` in my suggestion, so here's an even simpler version:

```
    if form.is_valid():
        form.save()
        return redirect('view')
messages.add_message(self.request, messages.ERROR, 'File upload error')
return render(self.request, 'air_quality/upload.html', {
    'form': SessionDataForm
})
```

If you still prefer the single return syntax, I would suggest using a variable name like `view_result` or something like that since `path` is a bit ambiguous and could be easily confused with `os.path` and `sys.path`.

woeip/apps/air_quality/migrations/0004_auto_20190613_0043.py
nbilenko: Agreed, I don't think it's a big deal since it's just a couple so far. I believe there's a method for squashing them but should be fine to do so at some point later when there are more accumulated!

woeip/apps/air_quality/views.py
TangoYankee: I'm convinced. I'll have it changed by this Hack Night.

I also think the linter configuration deserves its own issue.

nbilenko: > I'm convinced. I'll have it changed by this Hack Night.
> 
> I also think the linter configuration deserves its own issue.

SGTM for both!

Ethan-bradley:
Sounds good.

theecrit:
In the interest of expediency, I think we should move forward with this merge and open new Issues for the following items mentioned upthread:

**MVP Bugs**
- [ ] Attempting to upload a 2MB file results in failure

**MVP Improvements**
- [ ] All support code relating to the view, page, links, and url are meant to be changed or removed on the path to MVP
- [ ] Need to be careful that the session object exists and has the correct relationships, before populating the data table
- [ ] Enforce which file goes into which sessionData field in table
- [ ] UI enhancements

I'm not clear if there's an action item on the "Linter needs to be adjusted" notation.

If there are any concerns or objections, let's talk it out so we can unblock this. @nbilenko Have you had a chance to test this code at all? I'm hesitant to approve the PR myself, from a technical perspective. Lemme know what you think.

woeip/apps/air_quality/migrations/0002_sessiondata_upload_gps.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.1.7 on 2019-06-12 20:59
```
r-b-g-b, line 1: Ahhh migrations are a pain. Since we don't have any data in the db yet, I think we have the luxury of being able to blast all the old migration files and starting from a clean 0001_. I'd suggest that for the sake of simplicity, unless you know of a reason to track these db tweaks.

woeip/apps/air_quality/models.py
```diff
@@ -54,19 +54,20 @@ def __str__(self):
 
 
 class SessionData(TimeStampedModel):
-    """The raw data file generated during a session. Assumes one and only one file per sensor,
-    although multiple sensors can be linked to one session"""
-    upload = models.FileField(upload_to='session_data')
+    """The raw data files generated during a session. Assumes one general and one gps file.
+    Multiple sensors can be linked to one session"""
+    sensor_file = models.FileField(upload_to='sensor_files', default="")
```
r-b-g-b, line 59: At the risk of slowing things down, it's probably worth having one more chance to hash out the schema before we commit. I admit I didn't communicate very clearly when designing it, so any misunderstanding is my bad. But I think we can get something will a) work for most future data sources, and b) not be too cumbersome for the dustrak/gps data source that is the only one we currently have.

r-b-g-b (changes requested):

woeip/apps/air_quality/models.py
r-b-g-b: Here's what I'm thinking:
https://www.lucidchart.com/invitations/accept/5adb4594-1f61-4230-936f-32b522302382

I drew two cases. The top is the one we actually have, one device that records air quality and one that records GPS. Each produces their own raw data file. The bottom one is another (I think likely) case, one device that records air quality and GPS location and produces a single raw data file.

r-b-g-b: Also realize this is a bit of a detour from the current PR. Down to merge this since it contains a lot of awesome front end and form handling parts, and address the DB schema in a separate issue. In that case, we would still have to rewrite some of the DB parts of this PR, but could be worth it to keep moving forward.

theecrit:
Let’s move forward with this PR as-is (unless there functional fixes that
need to be included in order to safely merge). Then we can address the
database question separately, and with a more robust definition of the
problem.
-- 
. . . . . . . . . . . . . . . . . . . .
Visit me online at jessicasand.com

woeip/apps/air_quality/models.py
nbilenko: Agree with @r-b-g-b that this schema of the table is something we'll likely have to revisit later, and also that it makes sense to move forward with this PR as it stands now. 
It seems there's still a little ambiguity on what will make the most sense for the schema, and I think it will become much clearer as we start testing the data flow. So even if we revise the schema, committing to this solution now is a step forward.

nbilenko (approved):

woeip/apps/air_quality/models.py
r-b-g-b: Sweet, in that case I'll open a new issue and approve this request. I'll say don't worry about the migration comment I made earlier.

r-b-g-b (approved):
Moving database schema discussion to #115. Approved! Nice work @TangoYankee

Issue #122, opened 2019-07-29
Explore shifting the tech stack to take a microservices approach


theecrit:
**Is your feature request related to a problem? Please describe.**
Currently, the team is getting a lot of javascript/React developers who are interested in contributing. These folks often struggle with our tech stack, and it impacts retention. Additionally, tying the Django database directly into the front-end Zurb framework may cause scalability issues in the future [needs confirmation/clarification from @ifed3].

**Describe the solution you'd like**
The current proposal is to uncouple the front-end from the current Django back-end, and rebuild the former in React (javascript). The front-end will then rely on the Django (python)-driven API to make calls to/from the PostgreSQL database. See attached diagram and branch `ia-set-up-microservices`.

**Describe alternatives you've considered**
See attached diagram for existing solution.

**Additional context**
This issue adds a design dependency: we'll need to identify a front-end pattern library to use (or create/adapt one). This is currently being addressed in [Trello](https://trello.com/c/fGvq6ou6), and will be converted into a Github issue once an MVP pattern library is ready for implementation. 

**Labels**
- [x] Add a status label
  - `implementing`
- [x] Add additional labels as needed

### For Developers
**Affected Components**
List what pages or backend files will need to be added, removed, or changed

**Technical Resources**
TBD by @ifed3

**Design Documentation**
[WOAQ-Data-Flow-Diagram-v4.pdf](https://github.com/openoakland/woeip/files/3444251/WOAQ-Data-Flow-Diagram-v4.pdf)

**Testing Outline**
TBD by @ifed3

theecrit:
Closing as addressed by PR125.

PR #125, opened 2019-08-14
Move to a microservice architecture

ifed3:
---
name: Upgrade WOAQ infrastructure
about: #122
title: Set up microservices
labels: infrastructure, api, web, dataviz
reviewers: @r-b-g-b, @theecrit, @nbilenko, @TangoYankee 

---

## Checklist
- [X] Add description
- [X] Reference open issue pull request addresses
- [X] Pass linting check
- [X] Pass functional tests
- [X] Request code review
- [x] Merge own code

## Description
Issue: #122 

Thanks for your patience. This PR is finally ready.

* There are now five services in the WOAQ architecture: django rest api, react client, jupyter notebook for dataviz, postgis database, reverse proxy with automatically generated ssl certificates
* Services are no longer accessible by ports, instead they are defined by domains
    * Web interface is available on [lvh.me](http://lvh.me)
    * API is available on [api.lvh.me](http://api.lvh.me)
    * Data visualization notebook is available on [dataviz.lvh.me](http://dataviz.lvh.me)
* README has been cleaned up to make it easier to get up and running. Project overview and historical info have been moved to the About.md section
* Python dependency manager has been changed from pipenv to [pip-tools](https://github.com/jazzband/pip-tools). It's simpler for working with Docker environments. Read more [here](https://pythonspeed.com/articles/pipenv-docker/)
* Make commands that relate to docker have been removed due to refactored docker-compose files
and multiple services instead of a single one
    * Run `docker-compose up -d --build` instead of `make local.build` + `make local.up` to build and run all services
    * Run `docker-compose up -d --build service_name` to run a specific service
    * Run `docker-compose down` instead of `make local.down`
    * Run `docker-compose run service_name /bin/bash` instead of `make local.shell` to enter the shell of a specific service
* One thing to pay attention to is the change in directory structure. All services have their own directory. Make sure to `cd` into the correct directory if you want to run commands for a specific service

p.s there are still some outstanding issues relating to devops/infrastructure that will need to be completed in the near future: writing tests for the client service, creating a Heroku container configuration file, setting up the Heroku deploy process, pushing container images to Docker hub to enable faster CI builds


nbilenko:
Lots of awesome work here @ifed3. The infrastructure work looks great. The cleaner directory structure and separate Dockerfiles make it much easier to understand the codebase and to isolate changes related to different microservices.

However, there are changes that are out of scope of the issue, and this makes this PR quite large and more difficult to review. I would suggest retaining the work related to setting up microservices in this PR and separating out the other changes into different issues/PRs.

Some examples of changes I would suggest relocating:
- Python dependency management. I would be totally open to a change, but it's out of scope of this issue.
- Removal of `make` commands. Retaining `make` commands shouldn't impact anything about `docker-compose` commands since they are just aliases. It also seems unrelated to the issue.
- Major README cleanup. The cleanup looks great! But general cleanup for easier onboarding should be a separate issue. The changes in this PR should only be related to the microservice related changes.
- Adding a jupyter notebook. I would be concerned about running a jupyter notebook in production because of how computationally expensive it would be. Also, notebooks are great for interactive data analyses, but very bad for displaying pre-made visualizations. In any case, this change is out of scope of this issue.

Since I'm asking for a major PR change, I'll hold off on specific review of the infrastructure changes, but I'm looking forward to it - seems like a great way to move forward.

theecrit:
I agree with @nbilenko here; we really want to make sure our issues contain discrete implementations so work can be tracked more easily and discussions can remain manageable. The issue breakdown provided feels appropriate to me, too.

Thanks,
J.

ifed3:
Good catch @nbilenko with the dataviz service running in production. I've updated that.

I think it would be misguided to say that changes are out of scope of this refactor. They are interwoven in a manner that would require more work to separate them out in a new PR.

My aim was to build an architecture that can be run locally and on production without major configuration, so with microservices, three high level changes happened:
1. The only supported way to run the infrastructure is via docker, i.e you can't run it directly on your host machine
2. The only environment to run locally via docker is the dev/test environment. There will never be a need to run the prod or CI env locally
3. Subdirectories now exist thus each directory should only contain files relevant to its service

## Switch from pipenv to pip-tools
pipenv by default provides a virtual env, serves as both a dependency manager and installer. This is ideal if you want to run a python app on your host machine but is problematic for docker usage.

pipenv uses different install commands based on the env: `pipenv install --dev` vs. `pipenv install`. To solve this would require creating a different Dockerfile to run the api service in prod simply due to one command change. Or to create a conditional bash script to check the env and run the appropriate command.

Once there is no need to run the infra on your host machine, it actually becomes both unnecessary and more cumbersome to maintain support for pipenv.

## Makefile docker-compose aliases
Once the django app is moved to the api directory, it takes the `Makefile` with it since most of the commands are python/django related.

To keep the aliased docker commands would require creating a new Makefile within the root directory. The aliases were ideal when you could choose an env to run (local vs. production) but became redundant when you're limited to only run the local env

So you go from:
```bash
make %.up: ## Start the (local|production) Docker containers
    docker-compose -f docker-compose.yml -f docker-compose.$*.yml up -d
```
to: 
```bash
make local.up ## Start the local Docker containers
    docker-compose -f docker-compose.yml -f docker-compose.local.yml up -d
```

However by default, docker-compose searches for two config files (`docker-compose.yml, docker-compose.override.yml`) when running; once I renamed `docker-compose.local.yml` to `docker-compose.override.yml`, these commands all became equivalent!
```bash
make local.up
docker-compose -f docker-compose.yml -f docker-compose.override.yml up -d
docker-compose up -d
```

Thus all the aliases essentially became plain and easy to use docker-compose commands and getting familiar with docker-compose syntax will be far more important for one's coding chops than using arbitrary aliases especially for such basic commands. 

So the aliases became counterproductive both from a code functionality and developer usage standpoint with a microservices architecture

## README/contributing.md
These changes relate to the onboarding and use of the micoservices architecture. If I had left them as they were, half of it would be irrelevant and non-functional since it assumes just a django app rather than multiple apps.

## Holoviews
This is the change I agree is extraneous and probably out of scope. 

To your point, @nbilenko, Holoviews was built specifically to solve the problem of displaying pre-made visualizations. Read more [here](http://holoviews.org/user_guide/Plots_and_Renderers.html). It can output static displays in many formats such as `png`, `jpeg`, `svg`, `html` and dynamic displays as `html` + `js`. because it uses [Bokeh](https://bokeh.pydata.org/en/latest/) as its backend provider which is described as follows:

```
Bokeh is an interactive visualization library that targets modern web browsers for presentation. Its goal is to provide elegant, concise construction of versatile graphics, and to extend this capability with high-performance interactivity over very large or streaming datasets
```

----
Hope that explains some of the changes. I think this conversation would be best be expounded upon and resolved in person. See you all tonight!

nbilenko:
> I think it would be misguided to say that changes are out of scope of this refactor. They are interwoven in a manner that would require more work to separate them out in a new PR.

The interweaving of the changes is exactly what my feedback is about - the interweaving complicates both the review and keeping track the history of these changes. The different changes I mentioned are hard to delineate and to keep track of while reading the code. Also, discussion for all changes has to be maintained in a single comment thread. E.g. a chat about pros and cons of dataviz software sounds great, but an infrastructure PR comment thread is not where decisions should be made about what software the team will be using for data visualization.

I realize it would take some work for you to separate out the changes, but to read and understand these interwoven changes would instead place a substantial amount of work on the rest of the team. It would also likely to not lead to effective review - best practices tend to recommend limiting code changes to a few hundred lines because reviewers can't reliably detect problems with longer change lists.

Let's discuss more tonight.

ifed3:
hey all,

This is the new [infra document](https://user-images.githubusercontent.com/9001622/63615519-6b755d80-c59a-11e9-8fe8-2cab84eae2e7.jpg)

Here's a more thorough breakdown of the changes that happened within this PR. If you have any questions or any confusion, please let me know.

Docker
---------
- Remove all files related to AWS, Terraform, Elastic Beanstalk
- Move django files to api directory
- Create react client service within web directory
- Replace nginx service and configs with fully managed nginx-proxy service
- Use nginx-proxy to map ports to domains
- Set up CORS configuration for nginx-proxy so api domain can be accessed by any other domain
- Use nginx-proxy-letsencrypt-companion to automagically set up SSL certificates for domains
- Rename docker-compose.local.yml to docker-compose.override.yml
- Rename docker-compose.travis.yml to docker-compose.ci.yml to align with environment
- Create docker-compose.prod.yml to set up production environment
- Remove docker-compose commands from Makefile due to django directory change and renaming of local compose override file
- Update dockerignore files to minimize unncessary files being copied to containers
- Drop pipenv to avoid api Dockerfile duplication for multiple environments
- Remove pipenv environment variables from api Dockerfile
- Remove logging configuration in api Dockerfile and entrypoint script in favor of PYTHONUNBUFFERED env variable and nginx-proxy taking care of gunicorn logs 

CI (Travis)
----------
- Replace wait-for-it script with dockerize library to avoid mantainance issues
- Await database connection with dockerize wait command to prevent CI build/test failure
- Setup secure environment variables to be used on Travis CI
- Simplify CI build and test scripts
- Update code coverage script

Wiki
-----------
- Clean README to set focus on awareness of available services and getting started quickly
- Move project details and history to About section
- Update contributing section to remove instructions on running on host machine (now only supported via docker-compose)
- Update contributing section to remove details about logging in to django admin (not needed for api usage)
- Update contributing section to provide a clear step-list for getting from issue to pull request
- Update contributing section to provide a one-liner for testing the api service without using aliases

Python
---------
- Replace Pipfile/Pipfile.lock with requirements.in/requirements.txt 
- Set python version with runtime.txt since Pipfile no longer exists
- Fix failing test from get_target_data method due to new pandas lib version being timezone aware
- Fix failing test from test_joiner and load_dustrak methods due to complex newline split script

docker-compose.override.yml
```diff
@@ -0,0 +1,59 @@
...
+
+  api:
+    build:
+      args:
+        - PYTHON_ENV=dev
+    environment:
+      VIRTUAL_HOST: api.lvh.me
+      VIRTUAL_PORT: 8000
+      SECRET_KEY: secret_key
+      DEBUG: true
```
nbilenko: This has to be in quotes to get passed as an env variable to Python b/c otherwise yaml treats it as a boolean:
`true` => `'True'`
C.f. https://github.com/docker/compose/issues/1788

docker-compose.override.yml
```diff
@@ -0,0 +1,59 @@
...
+      DATABASE_URL: postgis://postgres:postgres@db:5432/woaq
+      ENABLE_LOGGING_TO_FILE: false
```
nbilenko: This has to be in quotes, same as above:
`false` => `'False'`

README.md
```diff
@@ -1,51 +1,50 @@
...
+  git clone https://github.com/openoakland/woeip.git
+  cd woeip
+  docker-compose up -d --build
+```
 
-### Data science/analysis options
-We use [Trello](https://trello.com/b/EBnxZHmx/west-oakland-air-quality) to document and discuss data decisions:
-- Review and add a comment to any yellow “Data Q” card currently sitting in the [Qs About Data Analysis](https://trello.com/b/EBnxZHmx/west-oakland-air-quality) column. Or;
-- Add a new yellow “Data Q” card to the [Qs About Data Analysis](https://trello.com/b/EBnxZHmx/west-oakland-air-quality) column for any new questions you think of that might be important to answer. 
+Browse the client at [lvh.me](http://lvh.me), ping the api at [api.lvh.me](http://api.lvh.me).
```
nbilenko, line 20: Suggestion: add a line about how to bring the service back down.

README.md
```diff
@@ -1,51 +1,50 @@
...
+Run the woaq infrastructure with [Docker Compose](https://docs.docker.com/compose/)
 
-### Design & research options
-We use [Trello](https://trello.com/b/EBnxZHmx/west-oakland-air-quality) for project management:
-- Review and add a comment to any light blue “Design & Research” card currently sitting in the [In QA](https://trello.com/b/EBnxZHmx/west-oakland-air-quality) column. Or;
-- Grab any card tagged with the light blue “Design & Research” label in the [Ready to Work On](https://trello.com/b/EBnxZHmx/west-oakland-air-quality) column and get started. 
+```bash
+  git clone https://github.com/openoakland/woeip.git
+  cd woeip
+  docker-compose up -d --build
```
nbilenko, line 17: Since the db name has changed, I had to create a new database `woaq`, so I would suggest adding that in the README as an instruction for initial setup. This was the way I did it:
```
docker exec -it woaq-db /bin/bash
psql -U postgres
CREATE DATABASE woaq;
```

nbilenko:
Thank you for the detailed breakdown of the changes and added documentation, Ife! This is excellent work. I added some inline comments about changes I had to make to get things running, but overall this looks great.

nbilenko (changes requested):
Thank you for the detailed breakdown of the changes and documentation, Ife! I added inline comments about changes I had to make to get things running. Overall, looks great to me.

docker-compose.yml
```diff
@@ -1,32 +1,29 @@
...
     volumes:
-      - ./conf/nginx.conf:/etc/nginx/conf.d/default.conf:ro
-      - logs:/var/log/nginx
-      - media:/public/media:ro
-      - static:/public/static:ro
-volumes:
-  logs:
-  media:
-  static:
+      - /var/run/docker.sock:/tmp/docker.sock:ro
```
r-b-g-b, line 29: Just curious: Is there a simple explanation for why we link the docker socket like this?

web/.dockerignore
```diff
@@ -0,0 +1,11 @@
+.dockerignore
+.env
+.env.example
```
r-b-g-b, line 3: Do we want to ignore `.env.example` too?

r-b-g-b (approved):
Very cool @ifed3 ! Thanks for this. I learned a lot from reviewing this -- [extending docker-compose files](https://docs.docker.com/compose/extends/), [dockerize](https://github.com/jwilder/dockerize), [pip-tools](https://github.com/jazzband/pip-tools).

I did my best to understand the web service, SSL, letsencrypt, etc., but can't claim to be very knowledgeable about that, so I'll trust that you made it work. At the very least, I was able to get it up and running using the nice docs you provided.

I second @nbilenko 's point about preferring smaller PRs, so let's keep that in mind in the future. Approved!

README.md
r-b-g-b: Talking to @ifed3 now about this. When I set this up fresh, I didn't have to create the database manually, but worth looking into what was different in your case. One solution would be to have an entrypoint script that does this.

r-b-g-b: Opened #126 to address.

Issue #136, opened 2019-09-17
When there is more than one mobile session collected on a single day, I want to know which one I’m viewing


theecrit:
## Description
In some cases, there will be multiple mobile sessions collected on any given day. When I’m viewing any of these sessions, I want to understand where the current one I’m viewing is in the sequence. Additionally, when the system displays a multiple sessions (e.g. in a grid or list view), it should display them in correct reverse chronological order. This requires each session to have a sequence attribute.

## Acceptance criteria
- [x] Database’s Collection entity has a Sequence attribute formatted as an integer.
- [x] *Sequence* attribute corresponds to the chronological order of all sessions collected within that calendar day.
- [ ] Sequence defines the display order of all maps.
- [ ] All open questions in Issue have been answered or moved into new Issues.

## Dependencies
Database entities are adjusted to accommodate above attributes.

## Related documentation
Wireframe: [WOAQ-Wires-ViewMaps-v2.pdf](https://drive.google.com/open?id=1kP2MIls9wsYGffVQS1F-9gQolBWv_YO2)

## Open questions
TBD

theecrit:
@nbilenko I'm wondering if this should be deleted, based on my Q in PR140:

> Sequence is appended to sessionName when displaying in the View Maps UI.

nbilenko:
> @nbilenko I'm wondering if this should be deleted, based on my Q in PR140:
> 
> > Sequence is appended to sessionName when displaying in the View Maps UI.

That'll be a front-end issue - PR140 addresses the backend capability for providing the sequence number.

Question re: 
> Sequence is displayed where relevant in the Upload UI (particularly with regards to the metadata confirmation view).

Is this needed? We discussed that if the user isn't uploading the sessions in order, this might be confusing. Additionally, this would be technically challenging - the data isn't yet saved to the database during the metadata verification step, and so the sequence number won't be yet available.

theecrit:
Yeah, I think you're right — I'll remove those, and if we discover during MVP testing that the user needs affordance for sequence in the upload flow, we can revisit. I do think it matters in the View Maps view, but PR140 enables the backend functionality for that.

nbilenko:
Sequence is now exposed as an API call `/collection/<id>/sequence` via https://github.com/openoakland/woeip/pull/152

theecrit:
@nbilenko Can we safely close this, given #152 is now merged? These criteria are still open, so I'm not clear if there's more work to be done:

>  Sequence defines the display order of all maps.
[...]

PR #140, opened 2019-09-18
Adds a counter property to Collection and removes route

nbilenko:
---
name: Counter property added to Collection, route removed
about: https://github.com/openoakland/woeip/issues/136
labels: MVP, backend, database
reviewers: ifed3, r-b-g-b, theecrit
---

## Checklist
- [x] Add description
- [x] Pass linting check
- [x] Pass functional tests
- [x] Request code review
- [ ] Merge own code

## Description
Issue: https://github.com/openoakland/woeip/issues/136

* Adds a counter property to the Collection object, which returns the counter of the collection on that date (e.g. if `my_collection` is the earliest collection on September 17th, `my_collection.counter` would be `0`). Also adds a string representation for the Collection object that prints the collection date and the counter in a 1-indexed format (e.g. `print(my_collection)` would output `'Collection 1 on Sep 17 2019'`).

* Removes route field from Collection (not used).


theecrit:
@nbilenko I'll defer to @ifed3 and/or @r-b-g-b for code review, but my only bit of feedback is on:

> Also adds a string representation for the Collection object that prints the collection date and the counter in a 1-indexed format (e.g. print(my_collection) would output 'Collection 1 on Sep 17 2019').

I'm wondering if this syntax eliminates flexibility on the front-end? In other words: what if we decide, when designing the View Map interface, that we want the page title to reflect the session date, but the sequence number to appear elsewhere (as in [the last page of this wireframe](https://drive.google.com/open?id=1kP2MIls9wsYGffVQS1F-9gQolBWv_YO2)).

Just want to make sure this output syntax doesn't handcuff us on the front-end design. Other than that, looking good to me!

nbilenko:
> @nbilenko I'll defer to @ifed3 and/or @r-b-g-b for code review, but my only bit of feedback is on:
> 
> > Also adds a string representation for the Collection object that prints the collection date and the counter in a 1-indexed format (e.g. print(my_collection) would output 'Collection 1 on Sep 17 2019').
> 
> I'm wondering if this syntax eliminates flexibility on the front-end? In other words: what if we decide, when designing the View Map interface, that we want the page title to reflect the session date, but the sequence number to appear elsewhere (as in [the last page of this wireframe](https://drive.google.com/open?id=1kP2MIls9wsYGffVQS1F-9gQolBWv_YO2)).
> 
> Just want to make sure this output syntax doesn't handcuff us on the front-end design. Other than that, looking good to me!

This printout is just for internal convenience - while it *could* in principle be used, the `counter` is what should be used programmatically on the front-end, and it's just a number.

theecrit:
Ok, that's what I wasn't sure of, thanks!

ifed3:
@nbilenko this is less of a counter and more of a sorted position index which is an intensive operation. To get information about 1 collection, you have to interact with every other collection. So if there are 100 collections, to find the position of each one you would have to repeat sorting the list of collections 100 times. As the number of collections in the db grows, this property computation will get slower. Additionally insert or delete operations can cause race conditions while the sorting is happening which will give positional skips or holes.

@theecrit For this reason and others below, I would suggest dropping "counters" entirely. Counters in this case act as a lossy proxy for sorting by time of day (ToD). This type of sorting is more ideal for client usage.

It would be simpler for the client to receive pertinent collections, sort by start timestamp and figure the position of the current collection in the array of sorted collections. This also allows the user to toggle UI parameters to figure out which sequence view they desire.

In terms of display, information is lost by using a position index instead of the actual ToD. Looking at [the last page of the map view wireframe](https://drive.google.com/open?id=1kP2MIls9wsYGffVQS1F-9gQolBWv_YO2), we can simply replace `Session 1` and `Session 2` with the actual time durations: `10:30am - 11:30 am` and `1:30 pm - 2:30pm`. That conveys more direct information at a quick glance than a numerical order since the user doesn't have to click into `Session 1` to figure out when the collection took place.

nbilenko:
> @nbilenko this is less of a counter and more of a sorted position index which is an intensive operation. To get information about 1 collection, you have to interact with every other collection. So if there are 100 collections, to find the position of each one you would have to repeat sorting the list of collections 100 times. As the number of collections in the db grows, this property computation will get slower. Additionally insert or delete operations can cause race conditions while the sorting is happening which will give positional skips or holes.

The property gets evaluated when called, and django uses lazy querying for querysets, so the SQL wouldn't run until the return. There also shouldn't be a race condition since the property is not saving anything to the db - it's only in python.

Re: sorting, I definitely agree that's a good thing to think through carefully. My thinking behind this was that since the sorting happens after filtering by date, it would only sort the number of sessions on that date, and that's unlikely to be prohibitively large. Also, we likely wouldn't be displaying the counter for large numbers of sessions at a time, and that limits the number of calls we would make to this property. Do you foresee a situation where this would be very big?

theecrit:
@ifed3 
> It would be simpler for the client to receive pertinent collections, sort by start timestamp and figure the position of the current collection in the array of sorted collections. This also allows the user to toggle UI parameters to figure out which sequence view they desire.

Yeah, I hear what you're saying. In some use cases, *sort order* is appropriate (e.g. displaying session thumbnails in order in a gallery view). But there's another use case, which is as an identifier (e.g. displaying the sequence # in the UI, as in "Sept. 28, session 2"). Would your proposed alternative allow for this additional use case?

nbilenko:
> > @nbilenko this is less of a counter and more of a sorted position index which is an intensive operation. To get information about 1 collection, you have to interact with every other collection. So if there are 100 collections, to find the position of each one you would have to repeat sorting the list of collections 100 times. As the number of collections in the db grows, this property computation will get slower. Additionally insert or delete operations can cause race conditions while the sorting is happening which will give positional skips or holes.
> 
> The property gets evaluated when called, and django uses lazy querying for querysets, so the SQL wouldn't run until the return. There also shouldn't be a race condition since the property is not saving anything to the db - it's only in python.
> 
> Re: sorting, I definitely agree that's a good thing to think through carefully. My thinking behind this was that since the sorting happens after filtering by date, it would only sort the number of sessions on that date, and that's unlikely to be prohibitively large. Also, we likely wouldn't be displaying the counter for large numbers of sessions at a time, and that limits the number of calls we would make to this property. Do you foresee a situation where this would be very big?

I actually just realized the sorting operation was unnecessary.

ifed3:
@theecrit Alright if we want to have the sequence, it's clear that only the frontend would ever use the sequence. Internal calls would simply filter by day and sort by timestamp.

With that, @nbilenko I would suggest moving the sequence computing from the model and only making it available to the frontend via an api request to `/collections/:id/sequence`

This is far more flexible since there can be multiple collection sequences to check. Here are just a few:

1. The order of this session in all sessions collected by anyone
2. The order of this session in all sessions collected by me
3. The order of this session in all sessions uploaded by me
4. The order of this session in all sessions collected and uploaded by me

If we let the sequence determination happen on the model, we can only account for # 1. However if we do it as an api request, then we can send additional data along to have a more specific request sequence determination

An example is as follows:

Request:

GET /collections/5/sequence

```js
params = {
  duration: 'today'
  collection: 'all', // one of ['all', 'collected', 'uploaded', 'collected_and_uploaded']
}
```

Response: 
```js
{
  order: 3, // 1-based index to allow for Session 3
  sequence: [10, 2, 5, 8, 1] // List of Collection ids in sequence for simple routing, i.e view Session 1 links to /collections/10
}
```

nbilenko:
> With that, @nbilenko I would suggest moving the sequence computing from the model and only making it available to the frontend via an api request to `/collections/:id/sequence`
> 
> This is far more flexible since there can be multiple collection sequences to check. Here are just a few:
> 
> 1. The order of this session in all sessions collected by anyone
> 2. The order of this session in all sessions collected by me
> 3. The order of this session in all sessions uploaded by me
> 4. The order of this session in all sessions collected and uploaded by me
> 
> If we let the sequence determination happen on the model, we can only account for # 1. However if we do it as an api request, then we can send additional data along to have a more specific request sequence determination
> 
> An example is as follows:
> 
> Request:
> 
> GET /collections/5/sequence
> 
> ```js
> params = {
>   duration: 'today'
>   collection: 'all', // one of ['all', 'collected', 'uploaded', 'collected_and_uploaded']
> }
> ```
> 
> Response:
> 
> ```js
> {
>   order: 3, // 1-based index to allow for Session 3
>   sequence: [10, 2, 5, 8, 1] // List of Collection ids in sequence for simple routing, i.e view Session 1 links to /collections/10
> }
> ```

Sure, the property can be made available via API request (I can make a new PR for that). Keeping it as a property would be better than calculating it within the request itself though, since then it will be available to all requests (and it seems like that may be needed). The cases you're describing can be covered by either adding arguments to the property or adding new properties - I didn't handle them in this PR since they aren't MVP.

nbilenko:
@ifed3 If I'm interpreting correctly, your general point is that this functionality could be implemented in the frontend altogether, which is true. The implementation I'm proposing uses a model property for two reasons: it might be used in multiple contexts by the frontend so it would be good to prevent that redundancy, and is a cheap attribute to compute on the backend. So overall, it would allow for more flexibility.

theecrit:
> @ifed3 If I'm interpreting correctly, your general point is that this functionality could be implemented in the frontend altogether, which is true. The implementation I'm proposing uses a model property for two reasons: it might be used in multiple contexts by the frontend so it would be good to prevent that redundancy, and is a cheap attribute to compute on the backend. So overall, it would allow for more flexibility.

I'm inclined to keep it as a property unless there's a strong reason _not_ to.


>  it's clear that only the frontend would ever use the sequence. 

Does this assumption need vetting?

ifed3:
@nbilenko Not necessarily. I'm fine with it being computed on the backend. My concern is specifically that the computation is happening in a property attribute.

Properties are simple and usually based only on direct manipulation of self attributes of the model. Right now this property requires non-self attributes resulting from an expensive database call to get all collections to filter from.

This means once a collection object is serialized it's gonna include the counter property whether or not the caller needs it, thus making unnecessary database calls.

A solution for avoiding redundancy is to make the counter an instance method. Then the way for the external world to interface that method would be through an api call of `/collections/:id/sequence`.

So if a frontend component only needs the collection info, it calls `/collections/:id`, if it only needs the collection sequence, it calls `/collections/:id/sequence`, and if it needs both, it calls both.

What I'm trying to avoid is always sending the sequence irrespective of whether a component needs it since then we're hitting the database for no reason.

nbilenko:
> @nbilenko Not necessarily. I'm fine with it being computed on the backend. My concern is specifically that the computation is happening in a property attribute.
> 
> Properties are simple and usually based only on direct manipulation of self attributes of the model. Right now this property requires non-self attributes resulting from an expensive database call to get all collections to filter from.
> 
> This means once a collection object is serialized it's gonna include the counter property whether or not the caller needs it, thus making unnecessary database calls.
> 
> A solution for avoiding redundancy is to make the counter an instance method. Then the way for the external world to interface that method would be through an api call of `/collections/:id/sequence`.
> 
> So if a frontend component only needs the collection info, it calls `/collections/:id`, if it only needs the collection sequence, it calls `/collections/:id/sequence`, and if it needs both, it calls both.
> 
> What I'm trying to avoid is always sending the sequence irrespective of whether a component needs it since then we're hitting the database for no reason.

I think it's a pretty cheap call (as discussed above), but I agree we might as well avoid the possibility of it becoming a concern. I converted the property to an instance method. I'll add the request in a new PR.

ifed3 (approved):
thanks for taking my concerns into consideration, appreciate it!

nbilenko:
> thanks for taking my concerns into consideration, appreciate it!

Thank you for raising them!!

Issue #145, opened 2019-10-09
As a system, I want to process the uploaded files sent by POST request


nbilenko:
Backend complement of https://github.com/openoakland/woeip/issues/129

The upload endpoint should accept files passed by the frontend using a POST request, initialize a Collection with associated CollectionFiles, save files to cloud, and return either a success or an error.

**Acceptance criteria**
- [x] Collection is initialized based on `starts_at` and `ends_at` metadata fields passed by the frontend.
- [x] CollectionFile objects associated with the Collection are initialized with file data from GPS and pollutant files passed by the frontend.
- [x] Files are uploaded to cloud.
- [x] Response is passed to frontend indicating success or failure.

**Dependencies and assumptions**
- [x] POST request contains a json object containing fileds `starts_at` and `ends_at` with datetime fields corresponding to beginning and end of the session and two files in the following format:
```
{"files": [{"file_name": "gps_file.log", "file_data": b"gpsfiledata"}, {"file_name": "dustrak_file.csv", "file_data": b"dustrakfiledata"}],
 "starts_at": <datetime_obj>,
 "ends_at": <datetime_obj>,
}
```
- [x] Frontend performs initial validation of the files (size, correct number of files, correct format, sensible dates and times for `starts_at` and `ends_at` fields).

nbilenko:
Closing via https://github.com/openoakland/woeip/pull/146

PR #146, opened 2019-10-09
Collection view and serializer for upload endpoint

nbilenko:
---
name: Collection view and serializer for upload endpoint
about: Upload
title: Collection view and serializer for upload endpoint
labels: MVP, backend, upload-view

---

## Checklist
- [x] Add description
- [x] Pass functional tests
- [x] Request code review
- [x] Merge own code

## Description
Issue: https://github.com/openoakland/woeip/issues/145

Adds create() method to the Collection view and serializer that processes a POST request to collection. Handles metadata object (containing `starts_at` and `ends_at` fields) and file data (passed as files in the request), initializes new Collection and associated CollectionFiles.


r-b-g-b (changes requested):
Probably worth including a test or two to this, but after that I think it's good to go!

api/woeip/apps/air_quality/serializers.py
```diff
@@ -56,6 +56,19 @@ class Meta:
...
+        collection = Collection.objects.create(
+            starts_at=validated_data.get('starts_at'),
+            ends_at=validated_data.get('ends_at'),
+        )
+        for file in files_data:
+            CollectionFile.objects.create(
+                collection=collection,
+                file=file,
+            )
+        return collection
```
ifed3: @nbilenko, what do you think about using a map function to write a bit cleaner code?

```python
class Meta:
    model = Collection
    fields = ["starts_at", "ends_at"]

def create(self, validated_data):
    collection = Collection.objects.create(**validated_data)
    files = self.context.get('view').request.FILES
    map(lambda file: self.__create_files(collection, file), files)
    return collection

def __create_files(self, collection, file):
    CollectionFile.objects.create(
        collection=collection,
        file=file
    )
```

api/woeip/apps/air_quality/views.py
```diff
@@ -38,6 +39,13 @@ def data(self, request, pk=None):
 
         return Response(serializer.data)
 
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
ifed3: @nbilenko, there is a cool serializer.is_valid use case with automatic validation http error response. So we can use that as a guard statement here

```python
def create(self, request, *args, **kwargs):
    serializer = self.get_serializer(data=request.data)
    serializer.is_valid(raise_exception=True)
    serializer.save()
    return Response(serializer.data, status=status.HTTP_201_CREATED)
```

ifed3 (approved):
@nbilenko the code is really clean 👍 and I want to move it forward! I added some suggestions on make the code more readable, so feel free to incorporate.

@r-b-g-b, at this stage, we can focus less on unit testing and do more with integration testing and documentation as testing. We can use [api documetation](https://www.django-rest-framework.org/topics/documenting-your-api/) to explore any endpoint route to test if the responses and errors are as expected

nbilenko:
Thanks for the reviews!

@r-b-g-b Thanks for asking for tests! I should have included them in the first place - writing them ended up uncovering a few different issues, so I'm glad you kept me honest and lets indeed continue to keep coverage thorough.

In addition to tests, I added temporary media file storage logic to the test set up - otherwise django creates a bunch of files in the media dir during testing that don't get removed.

@ifed3 that's a cool validation feature, updated. Please note that the upload endpoint expects the following from the frontend (as data json):
```
{"files": [{"file_name": "gps_file.log", "file_data": b"gpsfiledata"}, {"file_name": "dustrak_file.csv", "file_data": b"dustrakfiledata"}],
 "starts_at": <datetime_obj>,
 "ends_at": <datetime_obj>,
}
```

r-b-g-b (approved):
Looks good to me!

PR #155, opened 2019-11-17
Serve API static files with nginx-proxy

### LINKED ISSUE #149: As a developer, I want to include static files when developing locally
Currently the standard development mode of `docker-compose up --build` does not serve the static files for the API UI.

**To replicate**

```
docker system prune --volumes
docker-compose up --build
```

And visit http://api.lvh.me No static files, very :(

**Describe the solution you'd like**
One fix is to add the static urls to the `woeip/api/woeip/urls.py` with the following:
```
from django.conf import settings
from django.conf.urls.static import static
from .settings import DEBUG

if DEBUG:
    urlpatterns += static(
        settings.STATIC_URL, document_root=settings.STATIC_ROOT
    )
```

This (plus running `collectstatic` from the `woaq-api` container) works. However, there might be a better way to do it.

**Describe alternatives you've considered**
Our docker compose setup does have an nginx proxy that *I think* should handle serving the static files locally. It might be better to find out how to fix this than to force django to serve the files.

jayqi:
---
name: Pull Request
title: 'Serve API static files with nginx-proxy'
labels: ''
reviewers: ''

---

## Checklist
- [x] Add description
- [X] Reference open issue pull request addresses
- [X] Pass functional tests
- [X] Request code review
- [x] Merge own code

## Description
Issue: Resolves #149

This PR has the nginx-proxy serve the API app's static assets. Previously, static assets like css and js files were not served and not available.

### Key steps for solution

Basically, we need to run `python manage.py collectstatic` as part of the deployment to copy all static files to the location specified by `STATIC_ROOT` (`/api/public/static/`), and then make that available to the nginx-proxy. 

#### Django

This [Django documentation page](https://docs.djangoproject.com/en/2.2/howto/static-files/) is a helpful reference. The relevant items in `settings.py` for the API Django app are as follows:
- `STATICFILES_DIRS` -- following the code in `settings.py`, this is set to `/api/woeip/static/`. During development, Django puts static files here.
- `STATIC_ROOT` -- following the code in `settings.py`, this is set to `/api/public/static/`. Running Django's `python manage.py collectstatic` will copy all static files into this location.
- `STATIC_URL` -- this is the web location that Django will expect static files to be available at. This is set to `/static/`, which means in practice the files should be available at `api.lvh.me/static/`. 

I added `python manage.py collectstatic --no-input` to the `make migrate` command so it will run as part of setup. This puts the static files at the location specified by `STATIC_ROOT` (`/api/public/static/`).

#### nginx-proxy

I am following the steps in [this StackOverflow answer](https://stackoverflow.com/a/56053265). Because we are using `jwilder/nginx-proxy` which has the Django app and the nginx-proxy in separate Docker containers, this is a variation on the instructions for standard Djangon/nginx setups for solving this issue.

First, since they are two separate Docker containers, we'll need to have them share the static files through a common mounted volume. This is done in `docker-compose.yml`. We define a volume named `static`. For the `api` container, we point this volume to where we collected the static files. For the `proxy` container, we point this volume to a location where we will later tell nginx to serve the files. 

Next, we must set up nginx to know which directory path to bind to the `/static/` URL route. Normally, you'd put this in a configuration file `/etc/nginx/conf.d/default.conf` (inside the proxy container) which is what other Django/nginx tutorials will tell you to do. However, `jwilder/nginx-proxy` automatically creates that configuration for you, so we do it differently. We do this by adding configurations to `/etc/nginx/vhost.d/` (inside the proxy container). We do this by copying the configuration via the proxy's Dockerfile. (Note this is also what is being done for the CORS configuration.) The configuration and alias is defined in `nginx-proxy/API_location`, which binds the mounted directory to the `/static/` route. 

Note that the SO answer says to name the file `default_location`. This would make the `/static/` route available across all virtual hosts, i.e. static files will be at both `lvh.me/static/` and `api.lvh.me/static/`. We only need the API's static files for the api application, so we instead use the `$API_DOMAIN` Docker variable to copy to the file `api.lvh.me_location`. This functionality is described in jwilder/nginx-proxy's README [here](https://github.com/jwilder/nginx-proxy#per-virtual_host-location-configuration).


api/Makefile
```diff
@@ -28,3 +28,4 @@ detect_missing_migrations:  ## Determine if any apps are missing generated migra
 
 migrate: ## Apply database migrations
 	python manage.py migrate --noinput
+	python manage.py collectstatic --no-input
```
nbilenko, line 33: I would suggest adding a new alias for just collecting static files (e.g. `collectstatic`) since this doesn't really have to do with database migration.

nginx-proxy/API_location
```diff
@@ -0,0 +1,3 @@
+location /static/ {
```
r-b-g-b: Slight style preference for lowercased filename `api_location`.

api/Makefile
```diff
@@ -28,3 +28,4 @@ detect_missing_migrations:  ## Determine if any apps are missing generated migra
 
 migrate: ## Apply database migrations
 	python manage.py migrate --noinput
```
r-b-g-b: Going to make this `--no-input` (with dash) to match.

nginx-proxy/Dockerfile
```diff
@@ -1,3 +1,4 @@
 FROM jwilder/nginx-proxy
 ARG API_DOMAIN
-COPY cors.conf /etc/nginx/vhost.d/$API_DOMAIN
\ No newline at end of file
+COPY cors.conf /etc/nginx/vhost.d/$API_DOMAIN
+COPY API_location /etc/nginx/vhost.d/${API_DOMAIN}_location
```
r-b-g-b: And change this to `api_location`

api/Makefile
nbilenko: And then `make collectstatic` would also need to be added to `woeip/api/entrypoint.sh`.

nbilenko (approved):
One suggestion, otherwise lgtm

r-b-g-b (changes requested):

r-b-g-b (approved):
LGTM!

PR #192, opened 2020-04-08
File Uploads

joshling1919:
---
name: File Uploads
about: Request merging changes into the code base
title: ''
labels: ''
reviewers: ''

---

## Checklist
- [x ] Add description
- [x ] Reference open issue pull request addresses
- [ ] Pass functional tests
  - complete on the local machine with `make local.shell` `make test`
- [x ] Request code review
  - Please allow **36 hours** from opening a pull request before merging a pull request- even if it has already recieved an approving review.
- [ ] Address comments on code and resolve requested changes
- [ ] Merge own code

## Description
Issue: #129 #184

*Brief description of solution*

image

image

image

image

Issue #184, opened 2020-03-11
 As a user, I want to upload a pair of files via a drag-and-drop interface


theecrit:
*Description*
I have a CSV and log file from a single session, and I’d like to upload them to the server as seamlessly as possible via a drag-and-drop model.

**Note:** this issue is a counterpart to #129. Whichever is complete first should serve as the basis for the other.

*Acceptance criteria* 
- [x] Upload interface includes affordance for drag-and-drop per the attached prototype.
- [x] User can see a list of and remove selected files from the interface prior to confirming upload.
- [x] All open questions in Issue have been answered or moved into new Issues.

*Dependencies and assumptions*
- ~User must be authenticated (signed in).~ We'll add the account gateway once the front-end is in place.
- System uploads files to a server-side directory (same server that's hosting the database).
- Status notifications and progress indicators will be covered by a separate issue.
- View will be built in React.
- User can only upload one pair at a time.

*Related documentation*
- [Prototype: happy path](https://www.figma.com/proto/IZw8VvX1CsRc764CTvg4aRSU/WOAQ-Wireframes-Parent-File?node-id=340%3A5671&scaling=contain)
- [Prototype: error path example](https://www.figma.com/proto/IZw8VvX1CsRc764CTvg4aRSU/WOAQ-Wireframes-Parent-File?node-id=341%3A5663&scaling=contain) for reference only; not covered by this issue (Note: See #131 for list of error use cases)
- [WOAQ-Data-Flow-Diagram-v5.pdf](https://github.com/openoakland/woeip/files/4157171/WOAQ-Data-Flow-Diagram-v5.pdf)
- [WOAQ-Upload-v5-Wireframes.pdf](https://github.com/openoakland/woeip/files/3937179/WOAQ-Upload-v5-Wireframes.pdf)
- ~[WOAQ-Upload-v4](https://drive.google.com/open?id=1Al5ILibeITvFcentrr0ia9R5QLRs9pIq)~

*Questions?*
- For back-end questions, reach out to @jayqi
- For visual design questions, reach out to Niels or Kimberly Low in #WOAQ Slack channel.
- For UX/flow, content, or anything else, reach out to @theecrit

Issue #129, opened 2019-09-17
 As a user, I want to upload a pair of files via a file browse-and-select interface


theecrit:
*Description*
I have a CSV and log file from a single session, and I’d like to upload them to the server as seamlessly as possible using a file browse-and-select model per the attached happy path prototype (below).

**Note:** this issue is a counterpart to #184. Whichever is complete first should serve as the basis for the other.

*Acceptance criteria* 
- [x] Upload interface includes affordance for file browse-and-select per happy path prototype below.
- [x] User can see a list of and remove selected files from the interface prior to confirming upload.
- [ ] All open questions in Issue have been answered or moved into new Issues.

*Dependencies and assumptions*
- ~User must be authenticated (signed in).~ We'll add the account gateway once the front-end is in place.
- System uploads files to a server-side directory (same server that's hosting the database).
- Status notifications and progress indicators will be covered by a separate issue.
- View will be built in React.
- User can only upload one pair at a time.

*Related documentation*
- [Prototype: happy path](https://www.figma.com/proto/IZw8VvX1CsRc764CTvg4aRSU/WOAQ-Wireframes-Parent-File?node-id=340%3A5671&scaling=contain)
- [Prototype: error path example](https://www.figma.com/proto/IZw8VvX1CsRc764CTvg4aRSU/WOAQ-Wireframes-Parent-File?node-id=341%3A5663&scaling=contain) for reference only; not covered by this issue (Note: See #131 for list of error use cases)
- [WOAQ-Data-Flow-Diagram-v5.pdf](https://github.com/openoakland/woeip/files/4157171/WOAQ-Data-Flow-Diagram-v5.pdf)
- [WOAQ-Upload-v5-Wireframes.pdf](https://github.com/openoakland/woeip/files/3937179/WOAQ-Upload-v5-Wireframes.pdf)
- ~[WOAQ-Upload-v4](https://drive.google.com/open?id=1Al5ILibeITvFcentrr0ia9R5QLRs9pIq)~

*Questions?*
- For back-end questions, reach out to @jayqi
- For visual design questions, reach out to Niels or Kimberly Low in #WOAQ Slack channel.
- For UX/flow, content, or anything else, reach out to @theecrit

theecrit:
Update: currently waiting on upload API endpoint to wrap up: https://github.com/openoakland/woeip/issues/145

theecrit:
Updated wireframes based on more recent development decisions captured here: https://github.com/openoakland/woeip/issues/132

[...]

Issue #196, opened 2020-04-16
Front end tests: memory issues, config warnings, and reference errors


TangoYankee:
1. Memory Issue
    - From running `npm test`
    - `Error: watch /home/millertim108/woeip/web ENOSPC`
    - ENOSPC -> no space left on device
    - Another warning associated with the error `--watch is not supported without git/hg, please use --watchAll `
    - Investigating that error brought me to this [Create-React-App-Typescript Issue](https://github.com/wmonk/create-react-app-typescript/issues/282). 
    - This led me to line 32 of `node_modules/razzle/scripts/test`. It will watch tests that are not run by a CI or using coverage. I considered changing `--watch` to `--watchAll`. They may have solved one of the messages but I doubt it would resolve the error issue. The bigger problem is that changing an external module... that is also git-ignored... is dangerous and impractical.
   - Proposed solution is... to run coverage! It turns off the watching that threatens memory. It also allows us to track the coverage reports in CI, the same way we track the backend.
  - **Update:** adding a -watchAll flag in package.json works, too. I still prefer a --coverage flag.

2. Config Warning
   - From running `npm test` or `jest`
   - Warning `ts-jest[main] (WARN) Replace any occurrences of "ts-jest/dist/preprocessor.js" or  "<rootDir>/node_modules/ts-jest/preprocessor.js" in the 'transform' section of your Jest confi
g with just "ts-jest"`
   - Proposed solution... heed the warning!

3. Reference errors
   - From running `npm test` or `jest`
   - `Cannot find module 'routes' from 'App.tsx'`
   - There is also a reference to the Map component in the index file
   - The running application seems to find the references find. It's the tests that are baffled. 
   - ~~I think the easiest solution is to write the whole relative path, rather than the abbreviated versions.~~  
**Update:** @joshling1919 implemented [moduleNameMapper](https://github.com/openoakland/woeip/blob/fe30796446019aea786964a7b0bdf915390d7bec/web/package.json#L62) in an upcoming change. This appears to be a more scalable solution as changing file paths only requires updating one reference.  
**Second Update**: For TSLint, it doesn't seem to recognize both `dev` and `an array of whitelisted modules`  in [`no-implicit-dependencies`](https://palantir.github.io/tslint/rules/no-implicit-dependencies/). Consequently, my vote has switched back to writing relative paths.  

4. Linting Automatically
  - `tslint --project .`
  - `npm run lint`

joshling1919:
[...]

PR #198, opened 2020-04-18
Web reset + file uploads

joshling1919:
Moving updates from https://github.com/openoakland/woeip/pull/192 to a new branch since it was easier to resolve conflicts this way.

---
name: File Uploads
title: ''
labels: ''
reviewers: ''

---

## Checklist
- [x ] Add description
- [x ] Reference open issue pull request addresses
- [ ] Pass functional tests
- [x ] Request code review
- [ ] Merge own code

## Description
Issue: #129 #184

image

image

image

image


api/woeip/settings.py
```diff
@@ -67,6 +68,12 @@
 
 ROOT_URLCONF = "woeip.urls"
 
+CORS_ORIGIN_ALLOW_ALL = False
+
+CORS_ORIGIN_WHITELIST = (
+    'http://lvh.me',
```
TangoYankee: Will we need to change the whitelist and should it be an open issue?

web/src/components/App/index.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { Route, Switch } from 'react-router-dom'
+import { ThemeProvider } from 'emotion-theming'
+import { theme } from 'theme'
```
TangoYankee, line 4: Using just the module name works in running the application. However, the tests are not able to follow the paths. This applies not just to `theme` but all the custom module imports. I talk about it more in issue #196.

web/package.json
```diff
@@ -25,18 +32,17 @@
...
-    "tslint-react": "^4.0.0",
-    "typescript": "^3.5.2"
+    "tslint-plugin-prettier": "^2.3.0",
+    "tslint-react": "^4.2.0",
+    "typescript": "^3.8.3"
   },
   "jest": {
     "transform": {
-      "\\.(ts|tsx)$": "<rootDir>/node_modules/ts-jest/preprocessor.js",
+      "\\.(ts|tsx)$": "ts-jest",
```
TangoYankee: You beat me to the change in PR #197

web/src/components/Upload/index.tsx
```diff
@@ -0,0 +1,203 @@
...
+  const [files, setFiles] = useState<Array<FileWithPath>>([])
+  const [success, setSuccess] = useState<boolean>(false)
+  const { getRootProps, getInputProps } = useDropzone({
+    onDrop: (acceptedFiles: Array<FileWithPath>) => {
+      setFiles([...files, ...acceptedFiles])
+    },
+    multiple: true
+  })
+
+  const upload = e => {
```
TangoYankee: This is flagged in the tests as `Parameter 'e' implicitly has an 'any' type`

api/woeip/settings.py
```diff
@@ -30,7 +30,7 @@
     "django.contrib.gis",
 ]
 
-THIRD_PARTY_APPS = ["django_extensions", "rest_framework", "storages", "drf_yasg"]
+THIRD_PARTY_APPS = ["django_extensions", "rest_framework", "storages", "drf_yasg", "corsheaders"]
```
TangoYankee: Was this install prompted by a specific error? My understanding is that the [nginx server]( https://github.com/openoakland/woeip/blob/master/nginx-proxy/cors.conf) configures the cors headers. This should be how the maps component is able to make a [get request](https://github.com/openoakland/woeip/blob/994710f174c3412bcb66089452b6ea60bc5cdd87/web/src/components/Map/index.tsx#L49) to the pollutant values without Django CORS Header installed/configured.

joshling1919: ooh good call! 

so it was throwing this error before I added the cors config:
image

That was also back before I actually got the request to go through, so I had thought that was part of the issue, but it looks like it was actually just a red herring. Didn't realize that the nginx server was taking care of cors stuff for us so we might just need to tweak some config there to handle this error message (even though it doesn't actually cause the request to fail). I'll delete all the cors changes from django, thanks for catching this!

joshling1919:
@TangoYankee and I recently discussed how this PR has started to bloat past its initial scope of setting up a happy path upload of the CSV and Log files. We agreed that it would be best if I wrap this up at a point where users can browse and select or drag and drop files and get some helpful warning messages. After getting to that point, we can discuss with the team how to best split up the rest of this feature.

At this point, uploading the files work correctly, but there are hardcoded `starts_at` and `ends_at` values. I also had to update the `api/woeip/apps/air_quality_serializers.py` file so that the file was not nested under a dict. I tried really hard to make it work with the format that the API previously expected (file nested under dict), but I wasn't able to find a way that would prevent the File object itself from being stringified when it was being nested. As a result, those `serializers.py` tests are now failing.

@TangoYankee, you mentioned that maybe I should just remove that hitting of the endpoint altogether and revert any changes to `serializers.py`. Let me know if that's the path you ultimately prefer me taking, and I'd be happy to do it!

theecrit:
Wow, really looking forward to running through this on Tuesday.

nginx-proxy/cors.conf
```diff
@@ -1,4 +1,4 @@
-add_header 'Access-Control-Allow-Origin' '*';
+add_header 'Access-Control-Allow-Origin' '*' always;
```
joshling1919: this was needed to handle CORS issue whenever the API returned an error:
https://stackoverflow.com/questions/28160982/how-to-make-nginx-cors-configuration-work-when-server-returns-error

TangoYankee:
> @TangoYankee, you mentioned that maybe I should just remove that hitting of the endpoint altogether and revert any changes to `serializers.py`. Let me know if that's the path you ultimately prefer me taking, and I'd be happy to do it!

I think this would be good to discuss with the team at hack night.

api/.gitignore
```diff
@@ -62,6 +62,7 @@ coverage.xml
 # Django stuff:
 *.log
 !woeip/apps/air_quality/tests/data/gps.log
+!woeip/apps/air_quality/tests/data/random.log
```
TangoYankee: I think this change should be delayed until another PR

api/woeip/apps/air_quality/serializers.py
```diff
@@ -78,10 +78,8 @@ class Meta:
 
     def create(self, validated_data):
 
-        ## Basic validation that we have the right files
-
+        # Basic validation that we have the right files
```
TangoYankee: I think changes to this file should be delayed until another PR

api/woeip/apps/air_quality/tests/data/random.log
```diff
@@ -0,0 +1,184 @@
+$ADVER,3080,2.3
```
TangoYankee: I think this file shouldn't be included until the next PR

api/woeip/apps/air_quality/tests/test_api.py
```diff
@@ -14,6 +14,7 @@
```
TangoYankee: Changes to this file should wait to the next PR

nginx-proxy/cors.conf
TangoYankee: We can make look into making this part of the next version.

web/src/client.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { hydrate } from 'react-dom'
 import { BrowserRouter } from 'react-router-dom'
 
-import App from './App'
+import App from './components/App'
```
TangoYankee: the `./` shouldn't be necessary

web/src/components/App/index.test.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { render, screen } from '@testing-library/react'
 import '@testing-library/jest-dom/extend-expect'
 
-import App from './App'
+import App from '.'
 import { MemoryRouter } from 'react-router-dom'
 
 describe('<App />', () => {
```
TangoYankee, line 8: The tests for `heading` and `list` broke when the DOM changed. I replaced those two checks with [    expect(screen.getAllByRole('link')).toHaveLength(4)](https://github.com/openoakland/woeip/blob/7abc804a8bfbc6821a9c355698aeb552c3a6fc44/web/src/components/App/index.test.tsx#L15)

web/src/components/App/index.test.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { render, screen } from '@testing-library/react'
 import '@testing-library/jest-dom/extend-expect'
 
-import App from './App'
+import App from '.'
```
TangoYankee: request to read `components/App`

web/src/components/App/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+import Navbar from 'components/Navbar'
+
+import routes from 'routes'
+
+import 'semantic-ui-css/semantic.min.css'
```
TangoYankee, line 9: Importing this module propagates two changes for jest.
1. Create a mock style file as in [/__ mocks __/styleMock.js](https://github.com/openoakland/woeip/blob/7abc804a8bfbc6821a9c355698aeb552c3a6fc44/web/__mocks__/styleMock.js#L1)
2. reference said file in package.json moduleNameMapper
   - [`"^.+\\.(css|less)$": "<rootDir>/__mocks__/styleMock.js"`](https://github.com/openoakland/woeip/blob/7abc804a8bfbc6821a9c355698aeb552c3a6fc44/web/package.json#L69)
    - P.S: great call on keeping moduleNameMapper

web/src/server.tsx
```diff
@@ -3,7 +3,7 @@ import React from 'react'
 import { renderToString } from 'react-dom/server'
 import { StaticRouter } from 'react-router-dom'
 
-import App from './App'
+import App from './components/App'
```
TangoYankee: request remove `./`

web/src/components/Upload/util.ts
```diff
@@ -0,0 +1,43 @@
...
+      if (file.name.endsWith('.csv')) {
+        csvFile = file
+      }
+
+      if (file.name.endsWith('.log')) {
+        logFile = file
+      }
+    })
+
+    if (!logFile || !csvFile) {
```
TangoYankee, line 37: I like the solution of setting the files as they're found

TangoYankee (changes requested):
These changes should help limit the scope of the PR and also get it in sync with Master

TangoYankee (approved):

PR #214, opened 2020-05-05
read datetimes from files

### LINKED ISSUE #131: As a user, I want to understand if the files I uploaded aren't correct in some way
## Description
I would like the system to display visual feedback in real time that informs me if I've uploaded the wrong file type, too many files, etc—and how to correct the issue.

## Acceptance criteria
- [ ] Once user initiates upload, visual indicator remains in motion while upload is in progress.
  - [ ] API Upload endpoint: see #145 and #146
- [ ] If files are successfully uploaded, system displays a success banner on the Data Confirmation view.
  - [ ] API: success state on 200 status
- [ ] If file upload fails, system displays an appropriate error banner on the File Upload view.
  - [ ] *API end points needed*
- [ ] All banners push page content down, rather than obscuring content.
- [ ] Accessibility features have been included in code:
  - [ ] Error and warning banners have a role=”alert” (so they’re announced by assistive tech)
  - [ ] All other banners have a role=”status” to be read after critical alerts
  - [ ] All banners have an aria-live attribute
  - [ ] Banners use aria-describedby to describe their purpose to assistive technologies when they’re announced or receive focus (description = banner text per wireframe)
- [ ] All open questions in Issue have been answered or moved into new Issues.

## Dependencies and assumptions
- ~Sequencing order has been identified~ See: [Flow Diagram](https://github.com/openoakland/woeip/files/4172627/WOAQ-Data-Flow-Diagram-v5.pdf)
- Errors display on File Upload view in order to provide real-time feedback to user, per wires/prototypes below.
- Success displays on Data Confirmation view to reinforce transition to next step per wires/prototypes below.
- Message copy maintains consistent voice and style.
- Visual design maintains consistent styles

## Related documentation
- [Latest wires are v6](https://www.figma.com/proto/IZw8VvX1CsRc764CTvg4aRSU/WOAQ-Wireframes-Parent-File?node-id=341%3A5663&scaling=contain) (v5 versions can be run as prototype to see flow)
- [Prototype: error path example](https://www.figma.com/proto/IZw8VvX1CsRc764CTvg4aRSU/WOAQ-Wireframes-Parent-File?node-id=341%3A5663&scaling=contain)
- [WOAQ-Data-Flow-Diagram-v5.pdf](https://github.com/openoakland/woeip/files/4157171/WOAQ-Data-Flow-Diagram-v5.pdf)
- [WOAQ-Upload-v5-Wireframes.pdf](https://github.com/openoakland/woeip/files/3937173/WOAQ-Upload-v5-Wireframes.pdf) (under discussion)
- ~[WOAQ-Upload-v4](https://drive.google.com/open?id=1Al5ILibeITvFcentrr0ia9R5QLRs9pIq)~

## Open questions
- [ ] Need to define “success” (e.g. rote file upload vs. data validation)
- [ ] Need to confirm the following API end points: generic error, too many files, one file only, two CSVs or two logs, no date match

### LINKED ISSUE #208: Date & Time validation utility in front end for upload function
**Is your feature request related to a problem? Please describe.**
Delegated from Issue #132.
The API will only accept hits to the collections endpoint that have all the required fields, including start and end time. The front end is also expected to check whether the times from both files match.

**Describe the solution you'd like**
I would like functions in the front end that partially parse the gps and dustrak files to determine their start and end times. The GPS parser could look for the $GPRMC timestamp. As discussed below, we may only need the first timestamp for the GPS file. The Dustrak file could look for the Test Start Time/Date and Test Length fields. The check would be that they are with ~~10~~ two minutes of each other. 

**Describe alternatives you've considered**
We could check for dates and times listed in the file names. However, file names can be easily changed and broken. We could generate the time based on the current time. However, the current time may not have a string relationship with the time of colllection.

**Additional context**
Parsing these files will make the front end more bespoke to dustrak and gps files, making it less flexible to other file formats. However, it will make the experience more fluid and automatic within this use case.

**Labels**
- [x] Add a status label
  - `backlog`
  - `discussing`
  - `ready`
  - `implementing`
- [x] Add additional labels as needed

### For Developers
**Affected Components**
Most likely a utility ts file in the Upload Folder.

**Technical Resources**
There should probably two separate functions. Each function takes its respective file and return the required times or errors. Dustrak returns an object with a start and end time. GPS returns a start time.

This may be split into Two PRs.

Possible use of [Moment.js](https://momentjs.com/docs/)

**Design Documentation**
- [Database Schema as of 4/30/2020](https://user-images.githubusercontent.com/32755410/80737017-cb2bed00-8ac7-11ea-9282-5836501a5fb6.png)
- [Wireframes](https://www.figma.com/file/IZw8VvX1CsRc764CTvg4aRSU/WOAQ-Wireframes-Parent-File?node-id=340%3A5671l)

**Testing Outline**
Each function should test a success and an error path. There are testing files in the backend: [dustrak](https://github.com/openoakland/woeip/blob/master/api/woeip/apps/air_quality/tests/data/dustrak.csv) & [gps](https://github.com/openoakland/woeip/blob/master/api/woeip/apps/air_quality/tests/data/gps.log)

TangoYankee:
## Checklist
- [x] Add description
- [x] Pass functional tests
  - spin up docker container service `docker up -d --build`
  - run api tests `make test`
  - run front-end tests `npm test` or `jest`
  - lint `npm run lint` or `tslint -p .` with optional `--fix` 
- [x] Request code review
- [ ] Merge own code

## Description
Issues: 
- #208 Date and Time validation utility
- #131 Uploaded files aren't correct

## Future Issues
- Color Error/Warning message according to wireframes
- Test `validateFiles` function
- A faster and more robust way to extract the data from the files. The current process waits for both files to be fully extracted. Big O(n). ~~Then, it splits the resulting string by each new line Big O(n).~~ When looking at specific values, the process looks at a specific index. If the wrong value is at that index, the time will be invalid. If there is no value at that index. It will cause an error.   
Incremental improvements include only splitting until the [`nth`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split) newline character. **Update:** I implemented the `nth` split.  
We could also replace checks of a specific index to REGEX searches.  
Ultimately, I'm interested in swapping the full extraction to a [data stream](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) and optimizing from there. 

- ~~A way to run file extraction concurrently in the tests. The file processing for the tests is currently synchronous. Some of my test runs have taken 8s. This will scale poorly when a whole suite of tests are built.~~ Edit: Did that, too


web/src/components/Upload/util.ts
```diff
@@ -1,43 +1,132 @@
...
+    let csvFile: File | undefined
 
     files.forEach(file => {
-      if (file.name.endsWith('.csv')) {
+      if (file.type === 'text/csv') {
         csvFile = file
       }
 
-      if (file.name.endsWith('.log')) {
+      if (file.type === 'text/x-log') {
```
joshling1919: I'm on a mac and the `file.type` for my .log file shows up as an empty string for some reason:
image

As a result, I failed this validation with these two files: https://drive.google.com/drive/folders/1_cbfTu6-GiCgZzez7TCBef4dVv0HK2m3

When I previously researched (briefly) checking for file type, I think there are some inconsistencies when it comes to these checks, and we might only be able to check for file extensions.

web/src/components/Upload/util.ts
```diff
@@ -1,43 +1,132 @@
...
+import { DustrakMeta, ValidateMeta } from 'components/Upload/types'
 
-// export const parseDatesFromCsv = (file: FileWithPath): Array<string> => {}
+export const getGpsStart = (textLines: Array<string>): moment.Moment => {
+  for (const line of textLines) {
+    if (line.startsWith('$GPRMC')) {
+      const lineFields: Array<string> = line.split(',', 10)
+      const encodedTime: string = lineFields[1]
+      const encodedDate: string = lineFields[9]
+      const endTime: moment.Moment = moment.utc(
```
joshling1919: should this be "startTime" to match the function name?

web/src/components/Upload/util.ts
```diff
@@ -1,43 +1,132 @@
...
+  }
+  const validateData: ValidateMeta = {
+    message: '',
+    dustrakMeta: dustrakData
   }
 
-  if (files.length === 2) {
-    let logFile: FileWithPath | undefined
-    let csvFile: FileWithPath | undefined
+  if (files.length > 0 && files.length < 2) {
```
joshling1919: overall I think we might want to take a more "composable" approach to writing these series of validations and avoid if {} else {} statements if possible. I think to do that, it might make more sense to just early return as soon as a validation fails? 

That way if we ever need to add or remove validations down the road, it would be a simpler process of just deleting one conditional rather than having to restructure conditional logic.

web/src/components/Upload/util.ts
```diff
@@ -1,43 +1,132 @@
...
+        !gpsStart.isValid()
+      ) {
+        validateData.message =
+          'Files could not be uploaded. Try again or choose a different file.'
+      } else if (
+        dustrakStart.subtract(2, 'minutes') <= gpsStart &&
+        gpsStart <= dustrakStart.add(2, 'minutes')
+      ) {
+        validateData.message = ''
+        dustrakData.startDatetime = dustrakStart.format()
```
joshling1919: from an efficiency standpoint, it makes sense to just grab the dustrakData while we're validating it, but I think since this function's main purpose is to just validate, and it's already doing a lot, we might want to do this in a separate function, even if it comes at the expense of needing to call that parse function again.

web/src/components/Upload/util.ts
```diff
@@ -1,43 +1,132 @@
...
+    const startTime: string = textLines[6].split(',')[1]
+    const startDate: string = textLines[7].split(',')[1]
+    const startDatetime: moment.Moment = moment.tz(
+      `${startDate} ${startTime}`,
+      'MM-DD-YYYY hh:mm:ss a',
+      'America/Los_Angeles'
+    )
+    return startDatetime
+  } catch (e) {
+    if (e instanceof TypeError) {
```
joshling1919: can you add a comment on when/why a `TypeError` occurs here and why we want to return a moment object in that scenario?

joshling1919 (changes requested):
Awesome work so far! Just had a couple of suggestions!

web/src/components/Upload/util.ts
TangoYankee: **Reponse**
How about an `If`, instead?

**Context**
This was originally an `if` statement covering the scenario where the file text array isn't long enough to have values in that array position. It's a type error because the index out of range condition returns `undefined` and you can't `split type of undefined`. Rather than comment, I think it's better practice to cover the known error condition with an `if` statement- it's more expressive in the way it's written and prevents excessive errors being triggered. I'll push an update to `if`. But I'll welcome your thoughts on whether it's better to keep it as an error.

web/src/components/Upload/util.ts
TangoYankee: **Response**
I'll revert
**Context**
The `if/else`s are coupled with the idea of having an object that tracks the state of validated times- there's a single return that has the final state of the whole object. But, I'm reverting back to only returning a string and might as well revert to a `if condition then return message` paradigm.

web/src/components/Upload/util.ts
TangoYankee: **Response Action**
I'll revert
**Context**
Ugh, cross-system compatibility.

web/src/components/Upload/util.ts
TangoYankee: **Response**
To implement this, I refactored the `csvFile` and `logFile` assignment step into a dedicated function- `identFiles`

**Context**
Trying to stay DRY when finding the csv file again in the data assignment step

web/src/components/Upload/util.ts
```diff
@@ -1,43 +1,115 @@
...
-      if (file.name.endsWith('.csv')) {
-        csvFile = file
-      }
-
-      if (file.name.endsWith('.log')) {
-        logFile = file
-      }
-    })
-
+    const [logFile, csvFile] = identFiles(files)
```
nbilenko, line 83: High-level comment - sometimes the files are distinguished using terms 'csv' and 'log' (like here) and sometimes 'gps' and 'dustrak' (elsewhere). It would be better to use the same naming scheme and make it fairly use case agnostic (e.g. 'gps'/'sensor' or 'location'/'sensor').

web/src/components/Upload/util.ts
nbilenko: I happened upon this discussion last week: https://github.com/react-dropzone/react-dropzone/issues/276
It seems that MIME types are extremely unreliable across browsers and operating systems, supporting your conclusion that extensions sound like the best way to go.

web/src/components/Upload/util.ts
TangoYankee: ...yeah, I couldn't decide which one made more sense. I chose the worst path and used both.

web/src/components/Upload/index.tsx
```diff
@@ -108,21 +112,30 @@ const Upload: React.FunctionComponent = () => {
...
+    const csvTextSplit: Array<string> = csvText.split('\n', 10)
+    const dustrakStart: moment.Moment = getDustrakStart(csvTextSplit)
+    formData.append('starts_at', dustrakStart.format())
+    formData.append(
+      'ends_at',
+      getDustrakEnd(csvTextSplit, dustrakStart).format()
+    )
     formData.append('pollutant', '1')
     axios
       .post('http://api.lvh.me/collection', formData, {
```
joshling1919, line 141: this is unrelated to your PR, but it just occurred to me that we should probably move `setFiles([])` that's currently on line 154 to inside the success `then` block so that we don't clear their files until it has successfully hit the API. I'm sure lots will change soon with the upload logic, but just wanted to note that.

web/src/components/Upload/util.ts
joshling1919: sorry to clarify, @nbilenko, would you prefer that these files be called `gps` and `dustrak` to keep it consistent with the backend?

web/src/components/Upload/util.ts
```diff
@@ -1,43 +1,115 @@
...
+      csvFile.text()
+    ])
+    const csvTextLines: Array<string> = csvText.split('\n', 9)
+    const dustrakStart: moment.Moment = getDustrakStart(csvTextLines)
+    const dustrakEnd: moment.Moment = getDustrakEnd(csvTextLines, dustrakStart)
+    const logTextLines: Array<string> = logText.split('\n', 10)
+    const gpsStart: moment.Moment = getGpsStart(logTextLines)
+
+    if (
+      !dustrakStart.isValid() ||
```
joshling1919, line 99: to clarify, is the empty `moment('')` invalid? Just wanted to be sure that doesn't return the current date instead since I think `moment()` without anything passed into it does that.

joshling1919 (approved):
Had a couple of thoughts/questions, but overall this looks good! Nice work! 🎉

web/src/components/Upload/util.ts
nbilenko: I'd be fine with either, as long as it's consistent - going with `gps` and `dustrak` would indeed make it so we don't have to change it on the backend, so that seems easier!

TangoYankee: Issue #221

web/src/components/Upload/index.tsx
TangoYankee: Issue #222

web/src/components/Upload/util.ts
TangoYankee: Correct

Issue #132, opened 2019-09-17
As a user, I want to be able to correct inaccurate metadata in the files I just uploaded


theecrit:
## Description
Once I’ve uploaded session files, I want to know that the information was loaded into the system correctly and be able to make changes before saving tot the database.

## Acceptance criteria
- [x] System should automatically extract certain relevant metadata from the dataFrame or staging table and display them back to the user for confirmation:
  - [x] ~Collection date~ _Delegated to #208_ 
  - [x] ~Start time (defined as: earliest matched timestamp within a file pair)~ _Delegated to #208_
  - [x] Device make and model Issue #232 
  - [x] ~Pollutant and unit measure~(System current relies on at least one pollutant already existing in the database.)
- [x] Each input field is editable by user.
- [x] Selecting “Save”:
  - [x] Triggers the system to copy all data to main table. PR #226 
  - [x] Flushes all data from dataFrame/staging table. PR #226 
  - [x] Pushes user to View Map view (earliest session, if user has uploaded multiple sessions). Issue #159 
- [x] Selecting “Cancel”:
  - [x] Flushes all data from dataFrame/staging table. PR #226 
  - [x] Returns user to Add Data view. Issue #231 
- [ ] All open questions in Issue have been answered or moved into new Issues.

## Dependencies and assumptions
- Assumes upload flow has been implemented
- System will display dates and times in a 12-hour US format.
- Editing functionality and UI and will be inherited from identified design system (Semantic UI for the calendar component?).
- Requires a separate Issue for deleting/editing an entire data set or individual data points.
- View will be built in React

## Related documentation
Wireframes 
- [WOEIP Schema V2](https://user-images.githubusercontent.com/32755410/80738548-0cbd9780-8aca-11ea-9ea1-b612bfebb99f.png)

- [WOAQ-Upload-v5-Wireframes.pdf](https://github.com/openoakland/woeip/files/3937179/WOAQ-Upload-v5-Wireframes.pdf) (see page 9)
- [Wireframes Parent File](https://www.figma.com/file/IZw8VvX1CsRc764CTvg4aRSU/WOAQ-Wireframes-Parent-File?node-id=340%3A5671l) - going off of the "Upload Flow v5: Happy Path Prototype"

TangoYankee:
> System should automatically extract certain relevant metadata from the dataFrame or staging table and display them back to the user for confirmation: 
- Delegated to issue #208

TangoYankee:
[...]

PR #229, opened 2020-06-13
Ty/upload confirmation state

TangoYankee:
---
name: Pull Request
title: ''
labels: ''
reviewers: ''

---

## Checklist
- [x] Add description
- [ ] Reference open issue pull request addresses
- [ ] Pass functional tests
  - spin up docker container service `docker up -d --build`
  - run api tests `make test`
  - run front-end tests `npm test` or `jest`
  - lint `npm run lint` or `tslint -p .` with optional `--fix` 
- [ ] Request code review
- [ ] Merge own code

## Description
- The `upload button` appears whenever the files meet the correct conditions to be uploaded. If this is good enough to trust the appearance of an `upload button`, I figured it's good enough to make the `confirmation page` appear. Consequently, I the replaced `stageOne` state manager with a condition check that is a reciprocal of the `upload page`'s check. This seems like a simpler solution to me.

- The form data object is created based on the state of the files. Also, it's not built until the 'upload button' is pressed. I removed the `formdata` state hook and instead I'm using the files state hook and passing it to the `confirmation` component.

**Next steps**
- [x] Use the file state to fill out the confirmation page before upload.
- [x] Move the upload functionality from the `upload` to the `save` button.
- [x] Route to different pages on either cancel or success 

~~The code **should** be working. However, I'm getting a 400 error when I try to hit the API. This is a new error. Interestingly, it persists when I revert to master. I'm wiping my local copy and seeing if I can get master working again. After that, I'll check whether the 400 error still happens for me on this branch, too.  If someone wants to run this branch and master on their machine and then get back to me, I'd appreciate the feedback.~~  
**sigh** the database needs to be seeded with an initial pollutant value. Otherwise, it will wonder what you're on about with the random pollutant value your form is talking about. I didn't need to wipe and then rebuild my local codebase to troubleshoot that. However, that is how it happened chronologically.  
**close sighing**

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -2539,9 +2539,9 @@
       }
     },
     "@types/history": {
-      "version": "4.7.5",
-      "resolved": "https://registry.npmjs.org/@types/history/-/history-4.7.5.tgz",
-      "integrity": "sha512-wLD/Aq2VggCJXSjxEwrMafIP51Z+13H78nXIX0ABEuIGhmB5sNGbR113MOKo+yfw+RDo1ZU3DM6yfnnRF/+ouw==",
+      "version": "4.7.6",
+      "resolved": "https://registry.npmjs.org/@types/history/-/history-4.7.6.tgz",
+      "integrity": "sha512-GRTZLeLJ8ia00ZH8mxMO8t0aC9M1N9bN461Z2eaRurJo6Fpa+utgCwLzI4jQHcrdzuzp5WPN9jRwpsCQ1VhJ5w==",
       "dev": true
     },
     "@types/html-minifier-terser": {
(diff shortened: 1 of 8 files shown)


Ajay-Vishwanath (approved):
This PR looks great Tim. It was a smart move to pass just the `files` from the `upload` component and have the upload confirmation do the rest in order to avoid passing along the formdata. I think it should be fine to remove the `upload` button from the first component with the automatic rendering of the confirmation component, we can run it by Jess to make sure.

Couple Questions:

1. Do we want to append the device type from the drop-down when we are creating and sending over the `formData`, or is there more work that needs to be done on the backend//do we need to clarify what those device types are?

2. Are we still doing some sort of check to make sure the GPS and CSV start times align? 

And then a couple of points to address for a next PR (which I can take):

- Formatting of the collection date and start time to match Figma
-  When the page reroutes you back to `about`on clicking `cancel`, the navigation bar still has upload underlined/selected.
- Also noticed that when you upload non-csv and gps.log files, the page errors out and says 'Unhandled Rejection(TypeError): Cannot read property 'text' of undeined'. That error stems from the `upload confirmation` component which shouldn't be hit unless file length and error message checks are met which is interesting.

TangoYankee:
> Do we want to append the device type from the drop-down when we are creating and sending over the formData, or is there more work that needs to be done on the backend//do we need to clarify what those device types are?

My understanding is that we should pull the name of device from the file.  
Issue #232 (part of #132)

> Are we still doing some sort of check to make sure the GPS and CSV start times align?

Yes. That check is one of the ways an error message can be generated. This keeps the `upload` component from switching to the `confirm` component.  
No Issues here.

> Formatting of the collection date and start time to match Figma

Issue #233

> When the page reroutes you back to abouton clicking cancel, the navigation bar still has upload underlined/selected.

That's part of a bug that's been bothering me for awhile but I considered low priority- the underline change is based on user interactions with the header, rather than a reflection of the current route.  Issue #234

> Also noticed that when you upload non-csv and gps.log files, the page errors out and says 'Unhandled Rejection(TypeError): Cannot read property 'text' of undeined'. 

Started Issue #235 . Could you add steps to reproduce?

Issue #261, opened 2020-09-30
Pushing Users to Appropriate Date and Session in Map After Upload


Ajay-Vishwanath:
**Is your feature request related to a problem? Please describe.**
Once users upload data, they need to be pushed to the appropriate date and session on the map page. Right now, it defaults to today's session information. 

**Describe the solution you'd like**
On upload from upload confirmation, the user is pushed to the map display component where it shows the session information they just uploaded. 

**Labels**
- [x] Add a status label
  - `needs definition`
  - `needs help`
  - `ready to work on`
  - `in progress`
  - `backlog`
- [x] Add additional labels as needed

### For Developers
**Affected Components**
Upload Confirmation Page and Map Index/Control Panel Component.

**Technical Resources**
Add links to appropriate documentation. Describe possible code implementations.

**Design Documentation**
Wireframes or other design resources

**Testing Outline**
Describe the testing methods appropriate for this feature

Issue #260, opened 2020-09-30
Links to Open Session Files on the Map Page


Ajay-Vishwanath:
**Is your feature request related to a problem? Please describe.**
On the Map display mockup on figma, there is a link to 'view raw data' in the bottom of the session information. We need this to open the CSV and GPS log files that were uploaded for that session.

**Describe the solution you'd like**
Implementation of a 'view raw data' link that either downloads those 2 files or opens them up in separate tabs.

**Additional context**
The files are already held in the `currentCollection` object that exists in the ControlPanel state. Just a matter of creating a link that opens up those files. 

**Labels**
- [x] Add a status label
  - `needs definition`
  - `needs help`
  - `ready to work on`
  - `in progress`
  - `backlog`
- [x] Add additional labels as needed

### For Developers
**Affected Components**
ControlPanel component.

**Technical Resources**
Add links to appropriate documentation. Describe possible code implementations.

**Design Documentation**
Figma design found [here](https://www.figma.com/file/IZw8VvX1CsRc764CTvg4aRSU/WOAQ-Wireframes-Parent-File?node-id=753%3A813)

**Testing Outline**
Describe the testing methods appropriate for this feature

theecrit:
Quick question re: the UX flow here. "View raw data" implies the interface will display the data, whereas "Download raw data files" is more accurate for the behavior described in the issue. I don't have a strong opinion either way, other than to make sure the language matches the behavior. 

Viewing raw data in the interface probably requires more development overhead (New window/view or modal overlay? What kind of formatting and styling? Add a download link, too?). So maybe we default to "Download raw data files" and if users seem to want to view the data first, we can build that out later?

Ajay-Vishwanath:
> Quick question re: the UX flow here. "View raw data" implies the interface will display the data, whereas "Download raw data files" is more accurate for the behavior described in the issue. I don't have a strong opinion either way, other than to make sure the language matches the behavior.
> 
> Viewing raw data in the interface probably requires more development overhead (New window/view or modal overlay? What kind of formatting and styling? Add a download link, too?). So maybe we default to "Download raw data files" and if users seem to want to view the data first, we can build that out later?

@theecrit Definitely agree - we should make it say 'download raw data' and have it just do that for now.

theecrit:
Works for me!

Ajay-Vishwanath:
@theecrit did we get this fully working with download raw data taking us all the way there? Was 90% of the way there with my last PR

Issue #259, opened 2020-09-30
Device Name Saved to the Database


Ajay-Vishwanath:
## Description
WOEIP uses three DustTrack units, nicknamed `A`, `B`, and `C`. Currently, this comes into play in two places in the user flow:

1. On the Confirmation page (between Upload and View Map)
2. On the View Map page

We need to ensure the system is pulling and displaying the data accurately. Right now:

- [ ] The Confirmation page displays a proper device nickname, but I'm not sure if this is pulling from the actual database or if this is dummy text. We need to confirm if that's the case.
-  [ ] The View Map page is displaying the wrong string. That needs to be corrected.

Previously, we'd discussed that we would pre-populate the database's `Device` table with the three device nicknames (A, B, and C). To do this, we need to figure out which nickname applies to which serial number (the SNs are: 8530091203, 8530094612, and 8530100707). 

For the purpose of this issue, let's assign them arbitrarily and open a bug to correct the `name` fields once we have that info:

**Table: Device, Record 1** 
- Serial: 8530091203
- Name: DustTrak A

**Table: Device, Record 2** 
- Serial: 8530094612
- Name: DustTrak B
- 
**Table: Device, Record 3** 
- Serial: 8530100707
- Name: DustTrak C

### Confirmation page 
Confirm that the system:

- Identifies the `Serial Number` from the uploaded CSV file.
- Queries the `Device` table in the database for the `Serial Number`.
- Calls the associated `Name` field for that record in the "Device" field on the View Map page.

### View Map page
The View Map page displays a field for `Device`, which should display the nickname assigned to the particular DustTrak that collected the data. To get there, the system needs to:

- Query the proper table in the database to call the `name` field from the proper `Device` table record and display it in the "Device" field on the front-end.

## Original issue

**Is your feature request related to a problem? Please describe.**
We need the device type from the upload flow to be saved to the database in order to pull that information up on the session information section of the map display page.

**Describe the solution you'd like**
Append the device type to the `formData` in the upload confirmation page and save that information to the collection.

theecrit:
@Ajay-Vishwanath I think you're using _device type_ but really mean _device name_. 

- Device type = model (e.g. DustTrak), which isn't technically used anywhere in the [database](https://drive.google.com/file/d/1U5tIkROnRMZkQJXmoQan81WKCapi7fOy/view?usp=sharing).
[...]

Issue #258, opened 2020-09-30
Initial Error from Collection Date API Request


Ajay-Vishwanath:
**Describe the bug**
When changing the date on the Map Display page, the initial `collectionDateRequest` API call errors out which is printed to the console. This is not a huge deal as the next request is a successful 200-level request, but is worth fixing.

**To Reproduce**
1. Go to the Maps page
2. Open the Chrome console
3. Change the date 
4. See that cancel was logged in the console and a faulty network request was made

**Expected behavior**
A successful initial request.

**Screenshots**
Ajay can send you a screenshot if needed.

**Labels**
- [X] Add a status label
  - `needs definition`
  - `needs help`
  - `ready to work on`
  - `in progress`
  - `backlog`
- [X] Add a `bug` label
- [X] Add additional labels as needed

**Affected Components (For Developers)**
Map Index Component

kbrdsk:
This seems to have been fixed in PR#288, cancel is still logged to the console but no more faulty requests are being made.

theecrit:
Closing per @kbrdsk's observation above.

PR #266, opened 2020-11-16
Ajay map updates

Ajay-Vishwanath:
## Checklist
- [X] Add description
- [X] Reference the open issue that the pull request addresses
- [X] Pass code quality checks
  - lint `npm run lint-fix`
  - exit container as above
- [X] Request code review
- [x] Merge own code

## Description
Issue: #261 #257 #260

This pull request:

* Adds a loading indicator on the map when pulling the pollution data
* Pushes user to the appropriate map date after uploading their data
* Adds a 'Download Raw Data' button which almost gets us to download the data (more on that below).

## Next Steps

In order to get to Map UI MVP - there are 4 steps developers can take right now

1. Issue #259: Add 'Device Type' as a column to the `Collection` Table, have the frontend include that in the `formData` upon upload, and then pull that for the map display. 

2. Complete issue #260 - I've gotten the 'Download Raw Data' div in the `ControlPanel` component to bring us to the API endpoint which contains the relevant files - but can't key into `currentCollection.collection_files[0].file` even though I can see the `file` parameter in the `currentCollection.collection_files[0]` object. That code is found lines 53-57 of the `ControlPanel` index.tsx file.

3. Issue #258 - 'cancel' is being logged to the console every time we call the `getCollections` function in the map component. I don't know if we just need to remove that `console.log(error)` or if there's something not ideal happening.

4. Bug that appeared during demo - after we changed dates, the API call and pollution data parsing were not cancelled and the pins still appeared on dates where there were no collections. 

For future work to get to Map UI MVP - as a team we need to have discussions about:

1. Part of issue #257 - specifically about the second checkbox - pushing the user to the map immediately after clicking save. I don't think this will be ideal with what's happening under the hood - the `collections` for the specific dates are called as soon as the component is mounted, and if we push to the map page without the newest collection being successfully saved yet, I can see it not coming up. I think there should be some loading indicator on the upload page after clicking save that we could design out.

2. Issue #237 - developers can implement the 'air quality this session/today' display that appears on the [wireframe](https://www.figma.com/file/IZw8VvX1CsRc764CTvg4aRSU/WOAQ-Wireframes-Parent-File?node-id=753%3A0) once we figure out how we are going to calculate that. Currently we pull just collection pollution data one session at a time (was too slow and erroring out when we pulled all sessions from a day) - so we will have to think of something to get daily aggregate data and then processing that. Maybe upon save - the backend ends up parsing the files and saving an average AQ index per `Collection` instance, and then the frontend can average out the daily `Collection` AQ indexes after it retrieves the `Collections` set per day.

### DIFF
--- a/web/src/components/Map/index.tsx
+++ b/web/src/components/Map/index.tsx
@@ -8,68 +8,8 @@ import * as Elements from 'components/Map/elements'
 import ControlPanel from 'components/Map/ControlPanel'
 // import MapFilters from 'components/Map/ControlPanel'
 import Pin from 'components/Map/Pin'
-<<<<<<< HEAD
-import {
-  Pollutant,
-  PollutantValueResponse,
-  Viewport
-} from 'components/Map/types'
-import {
-  MAPBOX_ACCESS_TOKEN,
-  MAP_STYLE
-} from '../../constants'
-
-const StyledContainer = styled(Container)`
-  margin-top: 30px;
-`
-
-const ContentContainer = styled.div`
-  margin: 0px 130px 92px 130px;
-`
-
-const LowerHalfContainer = styled.div`
-  margin-top: 22px;
-  display: flex;
-`
-
-const MapContainer = styled.div`
-  height: 548px;
-  width: 65%;
-`
-
-const ControlPanelContainer = styled.div`
-  height: 548px;
-  width: 35%;
-  padding-left: 54px;
-`
-
-const FormMessage = styled.h3`
-  font-size: 1.5rem;
-`
-
-const initialViewport: Viewport = {
-  zoom: 12,
-  latitude: 37.812036,
-  longitude: -122.286675,
-  bearing: 0,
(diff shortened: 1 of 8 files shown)


web/src/components/Map/index.tsx
```diff
@@ -35,13 +40,25 @@ const Map: FunctionComponent<{}> = () => {
...
+          history.replace({ pathname: '/maps', state: {} })
+          setLoading(true)
           getPollutants(token, firstCollection.id)
-            .then(pollutants =>
-              pollutants
-                ? setPollutants(pollutants as Pollutant[])
-                : setPollutants([])
-            )
+            .then(pollutants => {
+              if (pollutants) {
```
TangoYankee, line 55: Aww, I liked the ternary operator :(

TangoYankee:
All the thoughts I have are already captured in the your next steps.

TangoYankee (approved):

PR #323, opened 2021-05-01
259 save device name

sribbleinc:
## Checklist
- [x] Add description
  - lint `npm run lint-fix`
  - exit container as above
- [ ] Request code review
- [ ] Merge own code

## Description
Issue: #259

Added the Device Names to the database and pull the corresponding name from the database by device serial number.


web/src/api.util.test.js
```diff
@@ -12,4 +12,8 @@ it("should create the url to get a collection by its ID", () => {
   expect(apiUrlCollectionById(1)).toEqual(
     "http://api.lvh.me/collection/1/data"
   );
+
+  it("should create the url to get all devices", () => {
```
TangoYankee: We'll need this test to be on its own level, outside of the previous test.

 The cheatsheet to run tests are in the  `.github/pull_request_template##Checklist` file and section. To run a specific test, instead of all of them: `run test src/[sub-path-to-file]/[file-name.test.js]`

web/src/components/upload/confirm.js
```diff
@@ -157,7 +157,7 @@ export const UploadConfirm = ({
               disabled
               name="device"
               id="device"
-              value={dustrakSerialToLabel[dustrakSerial]}
+              value={dustrakSerialToLabel}
```
TangoYankee: In line with changing this to be the whole device object, we'll want to change the property that we pass. We'll also want to change the prop types and JSDoc that explain what properties we're passing. Finally, we'll want to change the the way we use the property

web/src/components/upload/drop.utils.js
```diff
@@ -143,3 +145,11 @@ export const extractFileMetaContent = async (file, endLine = 10) => {
     return [""];
   }
 };
+
+/**
+ * Retrieve all of the devices
+ * @returns {Array<Device>}
+ */
+export const getDevices = async () => {
```
TangoYankee: As this is a function that we use in `index.js`, we'll want to create an `utils.js` to keep its utility functions. We can then place this function there. (The lack of sub-name for `utils.js` indicates that it belongs to the default file- `index.js`)

As it is a unit function, we can also add unit tests. However, we don't have the proper node modules for that, yet. So these tests can wait for another day.

web/src/components/upload/index.js
```diff
@@ -1,25 +1,32 @@
...
+        const deviceList = await getDevices();
+        setDevices(deviceList);
+      } catch(err) {
+        console.error(err);
+      }
+    })()
+  }, [dustrakSerial]);
+
+  const matchedDevice = devices.filter(device => device.serial === dustrakSerial);
+  const deviceName = matchedDevice.length ? matchedDevice[0].name : null;
```
TangoYankee: 1. As we expect the device to change when the dustrakSerial changes, we should track it with state.
2. As we iterate on the design, we'll want to save the device association in the database. In which case, we'll want the device id, too. We might as well keep the whole device object and pass it to the confirm component.
3. If we set the 'error' value to an empty version of its type, rather than 'null', then we can more reliably and concisely handle the error paths. With the change to using the device object, we would default to using an empty object.

web/src/components/upload/index.js
```diff
@@ -1,25 +1,32 @@
...
+      try {
+        const deviceList = await getDevices();
+        setDevices(deviceList);
+      } catch(err) {
+        console.error(err);
+      }
+    })()
+  }, [dustrakSerial]);
+
+  const matchedDevice = devices.filter(device => device.serial === dustrakSerial);
```
TangoYankee: 1. As we're interested in finding one exact match, we can use `.find`, instead of `.filter`. It's also more efficient, as it can stop once it finds a match- rather than needing to look at every element, like filter does.
2. The function can be moved into `useEffect` in order to set the state of the device at the point where it would change.
3. It is a pure function, which means we can move it to the utility functions. This allows us to a. separate the concerns of each file and b. easily apply a unit test to it

TangoYankee (commented):
These recommendations may feel abstract. To help make them more concrete,, there is an annotated branch with an example of the changes: https://github.com/openoakland/woeip/compare/ty/259-save-device-name-notes

TangoYankee (approved):
@sribbleinc Thank you for your contribution! Merge when you're ready :)

Issue #325, opened 2021-05-03
Add "beta" notice in the UI that data may not be accurate


theecrit:
## Description

If we're launching a live MVP site without account sign-in, we can't limit who uploads data. While it's highly unlikely anyone will upload non-WOEIP data given the fairly strict system requirements for file recognition, it would be irresponsible not to contextualize the data views. 

### Recommended solution

- [ ] Apply a yellow `dismissable block message` ([reference](https://semantic-ui.com/collections/message.html#dismissable-block)) to the View Maps and Upload Data pages.
- [ ] Message should be placed above global nav. 
- [ ] Message text: "WOAQ is currently in early release. You may encounter bugs and data may not be displayed accurately."

theecrit:
@kbrdsk Assigned this to you, as I see you opened #341 to address it. :)

kbrdsk:
> @kbrdsk Assigned this to you, as I see you opened #341 to address it. :)

Thanks! Sorry for not updating it

PR #341, opened 2021-06-29
add dismissable beta msg to map and upload pages

kbrdsk:
## Checklist
- [x] Add description
  - lint `npm run lint-fix`
  - exit container as above
- [ ] Request code review
- [ ] Merge own code

## Description
Issue: #325

Add a dismissable message component to ui and use it in a beta notice component that is displayed above the nav menu in the navigation component. 

The dismissable messge seems like it could be reusable but I put the styling in the same file as the component itself so it could easily be removed. Not sure if that's a good pattern.


web/src/components/nav/betaNotice.js
```diff
@@ -0,0 +1,20 @@
+import { Route } from "react-router-dom";
+import { DismissableMessage } from "../ui";
+
+const MESSAGESTYLE = {
+	background: "#fec61e",
+	marginTop: "1em",
+	display: "inline-block",
+	width: "100%",
+};
```
wendy-wm-wu: If we want to follow how the other files are structured, this should be moved to its own css file.

web/src/components/ui/index.js
```diff
@@ -59,6 +60,13 @@ export const SuccessMessage = ({ ...props }) => <Message success {...props} />;
 
 export const WarningMessage = ({ ...props }) => <Message warning {...props} />;
 
+export const DismissableMessage = ({ ...props }) => {
+  const [visible, setVisible] = useState(true);
+  const handleDismiss = () => setVisible(false);
+
+  return visible ? <Message onDismiss={handleDismiss} {...props} /> : null;
+};
+
```
wendy-wm-wu, line 69: Nice use of a custom hook here.

web/src/components/nav/betaNotice.js
```diff
@@ -0,0 +1,20 @@
...
+
+const MESSAGECONTENT =
+	"WOAQ is currently in early release. You may encounter bugs and data may not be displayed accurately.";
+
+export const BetaNotice = () => {
+	return (
+		<Route path={["/maps", "/upload"]}>
+			<DismissableMessage style={MESSAGESTYLE} content={MESSAGECONTENT} />
+		</Route>
+	);
```
wendy-wm-wu: I'm still unsure if this should be its own separate component or if it should live within one of the router files for encapsulation. I'll let @TangoYankee make a call on this.

web/src/components/nav/betaNotice.js
```diff
@@ -0,0 +1,20 @@
...
+import { DismissableMessage } from "../ui";
+
+const MESSAGESTYLE = {
+	background: "#fec61e",
+	marginTop: "1em",
+	display: "inline-block",
+	width: "100%",
+};
+
+const MESSAGECONTENT =
```
wendy-wm-wu: This should be written in camel case.

web/src/components/nav/betaNotice.js
```diff
@@ -0,0 +1,20 @@
+import { Route } from "react-router-dom";
+import { DismissableMessage } from "../ui";
+
+const MESSAGESTYLE = {
```
wendy-wm-wu: This should be written in camel case.

web/src/components/nav/betaNotice.js
TangoYankee: As it's a constant that is scoped to at least the whole file, I can see the argument for non-camelcase. Although, I would advocate for SCREAMING_SNAKE_CASE.

```suggestion
const MESSAGE_CONTENT =
```

web/src/components/nav/betaNotice.js
TangoYankee: I would agree. Perhaps we have a `betaNotice.css` file and the `DismissableMessage` be passed the class?

This would also avoid the question of whether `MESSAGESTYLE` should be camelcase.

web/src/components/nav/betaNotice.js
TangoYankee: I'm not sure what @wendy-wm-wu has in mind. It could be a great alternative. 

As I'm reading this implementation, I. Love. Small. Components. More generally, I like functions that only do one thing. They're often easier to understand, reuse, and debug. In this vein, I like how `BetaNotice` builds on `Route` and `DismissableMessage` before it is placed in `BrowserRouter`.

web/src/components/nav/betaNotice.js
kbrdsk: Sounds good, thanks for the feedback!

web/src/components/ui/index.js
kbrdsk: Thanks!

TangoYankee (approved):
Apologies for letting this fall off my radar.

Issue #295, opened 2021-03-07
Testing Switch


TangoYankee:
Tests for switch result in this error: 
```
    Cannot find module 'worker-loader!mapbox-gl/dist/mapbox-gl-csp-worker' from 'src/components/map/box.js'
```

## Acceptance Criteria
- [x] Resolve Error
- [x] Do not break production build of map that is using hack that is triggering error
- [x] Write unit tests for switch component

theecrit:
@TangoYankee Is this a critical bug required for go-live functionality? Can it be moved into either "Ready to work on" or the P2 post-MVP project backlog?

TangoYankee:
Yes, it can be moved to post MVP. It affects our ability to run unit tests. But, we're not really enforcing unit tests.

PR #360, opened 2021-08-29
295 testing switch

TangoYankee:
## Checklist
- [x] Add description
  - lint `npm run lint-fix`
  - exit container as above
- [x] Request code review
- [ ] Merge own code

## Description

These changes allow our tests to pass. They also make it easier to create new tests and integrate them in the CI process.

For this issue, I investigated the source of an error for tests that involve rendering the map. From this GitHub [mapbox-gl-js comment](https://github.com/mapbox/mapbox-gl-js/issues/10487#issuecomment-805106543), gl-js used in the "box" component has hard requirements for browser apis. Meanwhile, jest runs on nodejs. Consequently, we either need to either mock a browser or rely on end-to-end tests. As we already have end to end tests in progress, we can build those up. Until then, we mock out the rendering of mapbox.

We are also getting spammed with a warning for an error in the React and Drop component we are relying on. We don't maintain this component and can't change it to fix the error. But, we can reduce the number of times the error appears in the console during test runs. We do this by mocking the upload component when we don't really need to render it completely.

Finally, I deleted the switch tests. They were being mocked to the point of meaningless and the targeted functionality is covered by other test suites.

Issue: #295


ontoneio (approved):
Looks good to me!

Issue #296, opened 2021-03-07
Production and Development Styling do not match


TangoYankee:
# Description
Some Semantic UI React components look tiny when compared to their development version

## Acceptance Criteria
- [ ] Production styling matches development styling

theecrit:
@TangoYankee Is this a post-MVP improvement, do you think?

TangoYankee:
> @TangoYankee Is this a post-MVP improvement, do you think?

yes. It will likely be resolved when we start applying our custom designs

theecrit:
I'm not seeing this fixed in #379 but let me know if I'm missing something.

TangoYankee:
> I'm not seeing this fixed in #379 but let me know if I'm missing something.

It is certainly possible I missed something. However, not all of the environments would immediately see a difference. Which environments haven't updated?

Issue #377, opened 2021-10-20
Sticky footer not sticking


theecrit:
**Describe the bug**

**To Reproduce**
1. Go to /about or /maps
2. View footer links overlapping the page content

**Expected behavior**
Footer component should stick to the bottom of the browser window and resize when window size changes, while remaining stuck to bottom of window. There should be a comfortable margin or padding of white space between the top edge of the footer and the bottom border of the page content, per the wireframe:

<img width="581" alt="Screen Shot 2021-10-20 at 5 26 09 PM" src="https://user-images.githubusercontent.com/32755410/138174869-dc355dfd-1284-49e9-b459-07520abae1a9.png">

**Screenshots**

![Screenshot 2021-10-20 at 17-13-50 West Oakland Air Quality](https://user-images.githubusercontent.com/32755410/138174935-4e103d86-25ae-4652-952f-7c6e7e71d624.png)
![Screenshot 2021-10-20 at 17-22-11 West Oakland Air Quality](https://user-images.githubusercontent.com/32755410/138174938-df276de7-4641-4678-9a4b-dcd8e049b62a.png)

**Desktop (please complete the following information):**
 - OS: OS 10.15.7
 - Browser: Firefox 93.0

theecrit:
Fixed in #379

Issue #246, opened 2020-08-12
Develop About page content and add to site


theecrit:
### Description
We'll want to include an About page that tells the public what they need to know about the project.

### Potential content requirements
- [x] Who's involved in the project and their roles (at the org level)
- [x] General context for the the project exists
- [x] Explanation of why this project is important
- [ ] ~Description of project success metrics/impacts if available~ `Won't include`
- [x] Links to deeper dive for specialty audiences (e.g. how to get involved)

### Acceptance criteria
- [x] Content has sign-off by WOEIP
- [x] Typo-free
- [x] All pages built in front-end:
  - [x] About WOAQ
  - [ ] ~West Oakland's Air~ Moving to separate issue
  - [ ] ~Get Involved~ Moving to separate issue
- [x] Changes pushed to PR ~#307~ #379 

### Other considerations
- Should be co-created with WOEIP
- Related to #34

theecrit:
Current draft: https://docs.google.com/document/d/1iBKiQjxV_n5qQuHQLAfsX1r5mIDKn9-ggksfDwkkx4g/edit?usp=sharing

theecrit:
This is currently blocked as I'm unable to get Docker functioning on Mac Catalina ([known issue](https://github.com/docker/for-mac/issues/5037)).

theecrit:
Adding notes from @TangoYankee re: proper tests to run:

> npm run test is correct
> npm run lint-fix may actually be lint:fix
> I merged in my changes which includes a new testing module.
> 'npm i' should do the trick

theecrit:
Fixed in #379

PR #379, opened 2021-10-23
Ty about

TangoYankee:
The same as js-about. But, it reactifies the class names, creates a sticky footer, and removes the font size adjustment that was creating problems in the production build.

Addresses #377 and #246


theecrit:
@TangoYankee Does this fix the sticky footer across the site? We have a separate issue open for that, and I think it makes more sense to open a separate PR that fixes it globally, rather than requiring every new page to address it independently, no? Maybe I'm misunderstanding the fix you included, though.

theecrit:
Also, what does `reactifies the class names` mean? :)

TangoYankee:
> 
> 
> Also, what does `reactifies the class names` mean? :)

React uses `className` instead of `class`. We had `class` in a few places. I just made a quick update.

TangoYankee:
> 
> 
> TangoYankee Does this fix the sticky footer across the site? We have a separate issue open for that, and I think it makes more sense to open a separate PR that fixes it globally, rather than requiring every new page to address it independently, no? Maybe I'm misunderstanding the fix you included, though.

It's a global fix that would close the footer issue. (It's technically poor form to combine multiple Issues into a single PR. But, the footer looked so bad on the About page and it was so easy to fix globally that I threw it in as  a bonus.)

theecrit:
Okay sweet, that all makes sense. I've updated all the issues you didn't 😉. I suppose this means we can kill my PR? I'm just troubleshooting a build fail on oo.org and will try to review and merge this in the next day or so. I'd like to review it for practice. :)

TangoYankee:
> 
> 
> Okay sweet, that all makes sense. I've updated all the issues you didn't 😉. I suppose this means we can kill my PR? I'm just troubleshooting a build fail on oo.org and will try to review and merge this in the next day or so. I'd like to review it for practice. :)

sounds good. Oh! and the font-size css that you noticed was the source of  #296. So these changes also remove that css. We can also close #296 with this PR.

Issue #346, opened 2021-08-04
Create front-end Sign In page


theecrit:
## Description
As a user with an existing account, I need a place to sign into it using my user ID and password. 

### Assumptions
- The Upload page will serve as the default home page for woaq.org for MVP.
- When visiting the Upload page, system will detect if user is signed in or not and if not, will display sign-in form by default.
- This issue covers only the creation of the front-end sign-in form that displays on Upload screen if user is not signed in.
- If user is not signed in, global navigation will include a "Sign In" option that directs user to this issue's sign-in form (global nav is not covered in this issue).
- Front-end can accept any user/password combo until user account back-end exists.

### Open questions
- [ ] When clicking "Sign In" in global nav, should user be taken to a new screen, should they receive a pop-up/modal sign-in form, or should a new page section be revealed (e.g. "push-down")? (Note, this issue is _not_ dependent on answering this question.)

### Requirements
- [ ] Front-end implements the components described in Figma [wireframes](https://www.figma.com/file/eJSGJTwzNaaYBufgYxKFRk/WOAQ-User-Account-Management?node-id=113%3A268) and accompanying [prototype](https://www.figma.com/proto/eJSGJTwzNaaYBufgYxKFRk/WOAQ-User-Account-Management?node-id=0%3A1).
- [ ] User can enter user name.
- [ ] User can enter password (hidden by default).
- [ ] User can reveal password.
- [ ] User can sign in with correct combo.
- [ ] Validation messaging appears inline as described in Figma wires/protoype above.
- [ ] On sign-in, user is redirected to previous page they were on when beginning sign-in flow.
- [ ] Hash the passwords in database so they are not stored as plain text.

exchrotek:
Still need to resolve CORS error on the backend. Tim and Angela's web dev friends may be a good resource.

TangoYankee:
Partially addressed by #435

PR #384, opened 2021-12-01
Draft: 346: account management

wendy-wm-wu:
## Checklist
- [ ] Add description
- [ ] Reference the open issue that the pull request addresses
  - lint `npm run lint-fix`
  - exit container as above
- [ ] Request code review
- [ ] Merge own code

## Description
Issue: #346

This is a POC for account management. This ticket will register, log-in, load, and log-out a user; it also verifies authentication and generates a JWT token. 

The password reset feature is not complete and will be taken up in a different story. We also will include further CSS changes as a separate story.


TangoYankee:
@theecrit @wendy-wm-wu I wasn't able to access the Netlify logs directly. However, I did attempt to build the application locally and found the front end failed to compile. I imagine the local and Netlify builds are failing for similar reasons.

wendy-wm-wu:
Thanks, @TangoYankee for looking into this. I'll go take a look. I must have missed this since I didn't re-run after rebasing.

web/src/components/auth/utils.js
```diff
@@ -0,0 +1,194 @@
...
+    let isAuthenticated = false;
+    if (res && res.data && res.data.access) {
+      isAuthenticated = checkAuthenticated(res.data.access);
+    }
+    if (isAuthenticated) {
+      setAccessTokenOnSuccess(res.data.access);
+      loadUser(res.data.access);
+    }
+    return isAuthenticated;
+  } catch (err) {
```
TangoYankee, line 68: I recommend that the catch block also return a value, in order to keep the function's return type consistent. This also, applies to other utility functions with catch blocks

web/src/components/auth/utils.js
```diff
@@ -0,0 +1,194 @@
...
+  localStorage.removeItem("refresh");
+};
+
+/**
+ * Logs-in a user
+ * @param {string} email the user's email
+ * @param {string} password the user's password
+ * @returns {boolean}
+ */
+export const login = async (email, password) => {
```
TangoYankee: The primary purpose of this function seems to be `make an api call to retrieve an access token.` I think it would be more maintainable to keep this function limited to that function by:

1) Changing the structure of return data to an array of two strings. The calling function can handle the responses accordingly
   - With success: return the token in the first index and an empty string in the second index
   - With Failure: return an empty string in the first index and the error in the second index
 2) Renaming the function to `getJWT` (the "T"  means token)
 3) Removing other functions that also cause side effects, unrelated to directly getting the JWT

web/src/components/auth/utils.js
```diff
@@ -0,0 +1,194 @@
...
+      "Content-Type": "application/json",
+    },
+  };
+  const body = JSON.stringify({ email, password });
+
+  try {
+    const res = await axios.post(apiUrlCreateJWTToken(), body, options);
+    let isAuthenticated = false;
+    if (res && res.data && res.data.access) {
+      isAuthenticated = checkAuthenticated(res.data.access);
```
TangoYankee: What is the motivation for checking the authentication of a token immediately after we retrieved it? My understanding is that the data should be fresh from the api and can be validated without another api call.

web/src/components/auth/utils.js
```diff
@@ -0,0 +1,194 @@
...
+    if (isAuthenticated) {
+      setAccessTokenOnSuccess(res.data.access);
+      loadUser(res.data.access);
+    }
+    return isAuthenticated;
+  } catch (err) {
+    console.error("Logging-in was unsuccessful", err);
+    setTokensOnFailure();
+  }
+};
```
TangoYankee, line 72: `loadUser`, `setAccessTokenOnSuccess`, `setAccessTokenOnFailure`, and `checkAuthenticated` each has its own side effect. I recommended moving them from this function and up to the same level/location where `login` is being called.

web/src/components/auth/utils.js
```diff
@@ -0,0 +1,194 @@
...
+ */
+const setAccessTokenOnSuccess = (access) => {
+  localStorage.setItem("access", access);
+};
+
+/**
+ * Sets access token to null on failure in Local Storage
+ *
+ */
+const setTokensOnFailure = () => {
```
TangoYankee:
```suggestion
const removeTokens = () => {
```

With a more generic name, we can more comfortable reuse it in other places- not just when there is a failure.

web/src/components/auth/utils.js
```diff
@@ -0,0 +1,194 @@
...
+    await axios.get(apiUrlLoadUser(), options);
+  } catch (err) {
+    console.error("Loading user was unsuccessful", err);
+  }
+};
+
+/**
+ * Logs-out a user
+ */
+export const logout = () => {
```
TangoYankee:
```suggestion
```

We shouldn't need to worry about a catch block. If the token doesn't exist, then nothing will happen. We should be able to go without this `logout` function because the generic `removeTokens` should provide the same functionality.

web/src/components/auth/utils.js
```diff
@@ -0,0 +1,194 @@
...
+    }
+  } else {
+    console.error("Authentication failed: no access token");
+  }
+};
+
+/**
+ * Loads a user
+ * @param {string} token
+ * @returns {User}
```
TangoYankee, line 171: Where is the `return`?

web/src/components/auth/utils.js
```diff
@@ -0,0 +1,194 @@
...
+ * @typeof WebToken
+ * @property {string} refresh
+ * @property {string} access
+ */
+
+/**
+ * Stores access token in Local Storage
+ * @param {string} access the access token
+ */
+const setAccessTokenOnSuccess = (access) => {
```
TangoYankee:
```suggestion
const setAccessToken = (access) => {
```

I think making the name more generic will make it feel more natural to reuse in other places

web/src/components/auth/utils.js
```diff
@@ -0,0 +1,194 @@
...
+ * Shape of User stored in database
+ * @typeof User
+ * @property {string} first_name
+ * @property {string} last_name
+ * @property {string} email
+ * @property {int} id
+ */
+
+/**
+ * Shape of JWT Web Token pair
```
TangoYankee:
```suggestion
 * Shape of JSON Web Token pair
```

This is a nitpick, I know

web/src/components/auth/resetPasswordConfirm.js
```diff
@@ -0,0 +1,13 @@
+import { Container } from "semantic-ui-react";
```
TangoYankee: I know it's weird. But, convention within the app would have us import and re-export "Container" through the "ui index"  and then import it from `ui`. This applies even if we haven't modified the component. The idea is to filter all of our components through the Custom UI folder, to 
1) track them in a central place and
2) make them as agnostic as possible

Same suggestion applies to the imports in `web/src/components/auth/resetPassword.js`

web/src/components/nav/bar.js
```diff
@@ -35,4 +57,11 @@ const links = [
   { text: "About", route: "about" },
   { text: "Upload", route: "upload" },
   { text: "Maps", route: "maps" },
+  { text: (token ? "Logout" : "Login"), route: "Login"},
```
TangoYankee: It looks like this is duplicating the above list

web/src/components/nav/bar.js
```diff
@@ -1,7 +1,29 @@
 import { Link, NavLink, withRouter } from "react-router-dom";
 import { Container, Menu } from "../ui";
+import { logout } from "../../../src/components/auth/utils";
+
+let token = localStorage.getItem("access");
```
TangoYankee: This may be better served by making it part of the state of the navbar component. When we initialize the state, we can grab the initialize value of the token.

web/src/components/nav/switch.js
```diff
@@ -39,4 +44,29 @@ export const routes = [
     exact: true,
     component: About,
   },
+  {
+    path: "/login",
```
TangoYankee, line 48: We may want to adjust these routes. I'm thinking that we may want to preface the account-related accounts with `account`. We may also be able to combine workflows into a single route with a component that holds both states
For example:

- `/account/auth` Handles login and logout with an "accountAuthentication" component
- `/account/password` Handles the reset and confirmation within a  "accountPassword" component
  - uid and token are held in the component's state, avoiding the need to put it in the url
- `/account` Can grow into an account management console page. But for now, if the user does not have an account, they can create one here. It would combine "register" and "activate" into a single component.

TangoYankee:
@wendy-wm-wu , would you be open to splitting this into two separate PRs- one for the front end and another for the backend? It should make it simpler review each one individually.

wendy-wm-wu:
@TangoYankee - Sorry for the late reply. I keep forgetting to check my other email that's linked to this GitHub. Since this is a PR with so many commits between both Alana and myself, I think it would be easier to leave as is if that's okay. In the future, we won't have a large MR like this and will split into frontend/backend.

Also, thank you so much for reviewing and for your thoughtful comments. I'll take a look soon.

web/src/components/auth/resetPasswordConfirm.js
wendy-wm-wu: Thanks for the reminder!

web/src/components/auth/utils.js
wendy-wm-wu: Valid point - I referenced the Djoser account management flow from somewhere, but you're right - we don't need the additional check here.

web/src/components/auth/utils.js
wendy-wm-wu: Sure, I can do that. 
1. What is the reason for using an array vs. an object?

web/src/components/nav/switch.js
wendy-wm-wu: For "activate", the path needs to remain as `/activate/:uid/:token`, but we can include `account` in front. When a user creates an account, they click a url in their email that brings them to the "activate" page with the `uid` and `token`. This is how the two values are generated.

wendy-wm-wu: I'll address this comment/push up additional changes once we're on the same page about this one.

web/src/components/auth/utils.js
TangoYankee: An object would work, too. Under the hood, arrays are technically objects, anyway.

web/src/components/nav/switch.js
TangoYankee: Yeah, on second thought, I don't think I like my recommendation.

TangoYankee:
I received the following error when creating a user account. I sent a `POST` request at the `auth/user` route using the the Django UI. In order to make sure the latest changes did not cause the error, I reset to the previous commit and tried again. The error remained.

Screenshot 2022-01-19 at 17-54-06 SMTPAuthenticationError at auth users

wendy-wm-wu:
@TangoYankee - Did you check your email? I think you have to click the url provided in the email, which will take you back to the site. There should be a button (still requires some CSS changes) on the page that you'll then click to activate your account.

TangoYankee:
> @TangoYankee - Did you check your email? I think you have to click the url provided in the email, which will take you back to the site. There should be a button (still requires some CSS changes) on the page that you'll then click to activate your account.

Even if we are utilizing email as part of the account creation process, the initial request should not return an error.

sribbleinc:
@TangoYankee Both Wendy and I have never seen this error before. Would you be able to record yourself producing the error to send to us? Or let us know if you want to schedule a time to go over it together.

TangoYankee:
@wendy-wm-wu I was able to sign in successfully. However, after I signed in, attempting to upload or view data was prevented by 401 errors.

wendy-wm-wu:
@TangoYankee - I looked into the issue, and I think we need to send "JWT token" with every request that we make. This will also require some backend changes that I'll need to dig into.

TangoYankee:
@wendy-wm-wu The upload page is working. It seems the map page needs the same treatment.

sribbleinc:
Hey @TangoYankee Wendy and I have found an issue with token persistence that we need to look into. We'll let you know when you can review again.

wendy-wm-wu:
@TangoYankee - Oops, that MR was not ready for review yet. I pushed up the initial change, so Alana could work on it. I've been using my company laptop to develop, and last week, I had to remove Docker desktop since it won't be free for commercial use anymore. Unfortunately, the Docker alternatives that I looked into do not support `docker-compose`. I'm going to try to set this up on my very slow personal computer, so I can start developing again. Like Alana mentioned, we'll let you know when this MR will be ready for review. Thanks for your patience.

Issue #395, opened 2022-02-06
As a user, I want to be able to see what dates in the calendar have data.


gsehrlich:
**Is your feature request related to a problem? Please describe.**
After uploading some data, if I close the page and return to it, I can't see which dates have data without opening each date one at a time.

**Describe the solution you'd like**
The calendar should have a marker or color coding for dates that have data. If there were some way to zoom out to a twelve-month or multi-year view, that might help the user scan for dates that have data.

**Describe alternatives you've considered**
A list on the side of dates for which data has been uploaded most recently.

**Labels**
  -  `component: front-end`
  - `feature: map`
  - `milestone: MVP`

- [ ] Get rid of "Today" button element
- [ ] Change background of date in the month to white

mnorelli:
Love this idea.  And... I would find it helpful that when clicking Save after uploading a pair of collection files, the Maps page opens to the date collected of those files.  Would this necessitate a new issue?

theecrit:
@mnorelli
> I would find it helpful that when clicking Save after uploading a pair of collection files, the Maps page opens to the date collected of those files. Would this necessitate a new issue?

It does this already! :D Let me know if it's not working for you and we can open a bug issue.

theecrit:
I think this is a really useful story to add (I could've sworn it used to do this, but I just tested it and sure enough it does not). 

It also raises a larger question around how we handle time. We've been prioritizing the data collector audience, which has resulted in us taking a very session-specific view of the data. But residents (and others) may want a view of AQ over time, or in aggregate. Example stories:

- As a data collector, I want to see the session I just uploaded. (Current experience)
- As a data collector, I want to see all the sessions I've uploaded in aggregate.
- As a resident or data collector, I want to see all uploaded data from everyone displayed in aggregate.
- As a resident or data collector, I want to see data from a specific range of dates.

We could create separate issues for each of these, but I feel like we should step back and do some design exploration around these together, so we can consider a more holistic approach. Our UX folks currently are focusing on mobile responsiveness but maybe we can capture these in a single design issue for now, for when we're ready to tackle that? And folks can comment on that issue with ideas/thoughts in the meantime.

mnorelli:
Glad you have the same experience <whew!>
[...]

PR #434, opened 2022-03-25
395 datepicker data

gsehrlich:
## Checklist
- [x] Add description
  - lint `npm run lint-fix`
  - exit container as above
- [x] Request code review
- [x] Merge own code

## Description
Issue: #395 

I'm an experienced developer of individual Python code for scientific applications, but I'm totally new to web development, Javascript, and collaborating on a codebase. Please feel free to offer any guidance, and thanks for bearing with me!

When the page loads, the map menu will now fetch data from the API to get a list of dates with at least one collection. The date is then automatically set to the most recent date with data. In the date picker, the dates with data are highlighted in blue and the dates without data are disabled so that they cannot be clicked.

- In /web/src/components/map:
  - Wrote new api call in utils.js and refactored similar calls
  - Added effect hook in index.js
  - Updated menu.js to use the data
  - Changed style in menu.css
  - Wrote tests for new behavior in utils.test.js and menu.test.js
- In api/woeip/apps/air_quality/:
  - Handled new api call in views.py
  - Wrote tests for new behavior in tests/test_api.py

Questions:
- [x] In views.py, is it okay to use `queryset.distinct`? All that's need is a list of distinct dates, but that method includes the unneeded columns of the table too (which isn't very much data). When I tried using `queryset.values_list` to omit those columns, the serializer threw an error.
- [x] I'm not sure how to test the desired behavior of automatically setting the date. That code is in the new effect hook in index.js.
- [x] In index.js, the linter complained that I check `location?.state?.date` but have no corresponding state variable in the dependency array, so I added `location` to it. Is that necessary? It shouldn't change after the page loads, right?
- [x] In menu.test.js, the linter complained that I wasn't using the data from July 19th that I added to correspond to the new date I added to `setOfDates`. I pacified it by putting all the collections data in a single object. Is there a better way of handling that? Should I omit the data entirely? It's true that it's unneeded, but it feels dangerous to omit it when `setOfDates` has that date.


web/src/components/map/index.js
```diff
@@ -46,10 +48,33 @@ export const Map = () => {
...
+   * Load a Set of all dates with at least one collection
+   * @modifies {api} calls to get one collection per date that has any
+   * @modifies {setOfDates}
+   * @modifies {mapDate}
+   * @returns {axios.CancelToken.source().cancel} cancel the api call
+   */
+  useEffect(() => {
+    const source = axios.CancelToken.source();
+    (async () => {
+      const { listOfDates, thrownCode } = await getAllDates(source);
```
TangoYankee: The cool kids indicate lists by making the variable name plural. (I understand that you're de-structuring an object and the first place to change it is actually the return object for the function).

```suggestion
      const { allDates, thrownCode } = await getAllDates(source);
```

web/src/components/map/index.js
```diff
@@ -29,11 +30,12 @@ import { Grid } from "../ui";
  */
 export const Map = () => {
   const location = useLocation(); // location for the url
-  const initialDate = moment(location?.state?.date) || moment(); // Date either from upload or current day
+  const initialDate = moment(location?.state?.date); // Date either from upload or current day (moment(undefined) gives now)
   const [mapDate, setMapDate] = useState(initialDate);
   const [formattedDate, setFormattedDate] = useState(
     initialDate.format("YYYY-MM-DD")
   );
+  const [setOfDates, setSetOfDates] = useState(() => new Set()); // Lazy initial render
```
TangoYankee: My personal choice for sets is to put `unique` in the name. Otherwise, we can end up in situations where we're saying `setSet...`

```suggestion
  const [allDatesUnique, setAllDatesUnique] = useState(() => new Set()); // Lazy initial render
```

TangoYankee: TIL about lazy initial states. I have a few situations at work where this will come in handy.

TangoYankee:
> In views.py, is it okay to use queryset.distinct? 

For a short term solution, I'm not too concerned. It does bloat the http response. It also pollutes the browser application memory. But, we only make the request once on render. Also, we immediately discard the bloat and the garbage collector should get to it before it causes problems. In the long term, I think there should be a way to access start dates directly. I'll write more about that on the original issue.

> I'm not sure how to test the desired behavior of automatically setting the date

Yeah, this is pretty complex. As a rough outline, I would make a mock request that returns a collection on a specific date. Await that response. Then, the data picker should display the start date of that mocked collection.

> In index.js, the linter complained that I check location?.state?.date

React dependency arrays can be a nightmare. Your misgiving are warranted. It may cause issues in the future, if the code gets a feature that updates the location for the component. But for now, it's fine the way it is.

> In menu.test.js, the linter complained that I wasn't using the data from July 19th

I think the law of diminishing returns is kicking in. It's probably not worth worrying much more about. Perhaps leave a comment providing background info for the next developer?

web/src/components/map/menu.css
```diff
@@ -2,3 +2,50 @@
...
+  background-color: royalblue;
+  color: white;
+}
+
+.clndr-days {
+  background-color: rgba(255, 255, 255, 1);
+  border: 1px solid white; /* hack to get rid of border */
+}
+
+/* Report tonight:
```
TangoYankee: Any outstanding questions we can move into follow-on issues and discussions. For now, I think we can remove this report and call the code good to merge.

TangoYankee (approved):

web/src/components/map/index.js
gsehrlich: Thanks! I'll change it in index.js and in the called function in utils.js, as well as utils.test.js.

web/src/components/map/index.js
gsehrlich: I was wondering if there was a better convention. Will change in index.js, menu.js, and menu.test.js.

web/src/components/map/menu.css
gsehrlich: Whoops, that wasn't supposed to be there. Thanks.

gsehrlich:
> > In views.py, is it okay to use queryset.distinct?
> 
> I'll write more about that on the original issue.

Thanks. This was what originally thinking we should do, but I agree that it's a longer-term issue.

> 
> > I'm not sure how to test the desired behavior of automatically setting the date
> 
> Yeah, this is pretty complex. As a rough outline, I would make a mock request that returns a collection on a specific date. Await that response. Then, the data picker should display the start date of that mocked collection.

This is beyond my capacity right now, so I think I'll leave this for a future update. Should I open a new issue for this?

Issue #464, opened 2022-05-04
Different PM2.5 units being used interchangeably


gsehrlich:
See issue #237.

Per a Slack conversation (edited for clarity and correctness):

There are not two but three different kinds of units under discussion here:
* PM2.5 **mass concentration in milligrams per cubic meter** (what the DusTrak records) or **micrograms per cubic meter** (what the EPA uses in their AQI calculation table).
* PM2.5 **particle count in PPM (parts per million)**, which appears in the wireframe and in (incorrectly) in the "archived" portion of the OP in reference to what the DusTrak records. The difference is that if the particles get more massive (while staying smaller than the 2.5 micron maximum diameter to qualify) but not more numerous, the PM2.5 mass concentration goes up but the PM2.5 particle count/concentration stays the same. PM2.5 particle count and PPM are not relevant to WOAQ and should be discontinued.
* **AQI**, which specifies the colors that WOEIP wants used to contextualize the numbers. Based on the 24-hour average, each pollutant gets a number linearly in between the break points in the table, and the largest of those numbers becomes the AQI with the corresponding pollutant as the "pollutant of concern". So the colors essentially say "this is what the AQI would be if PM2.5 were the worst pollutant and it had this value averaged over the last 24 hours."

We'll use the format "## PM2.5," where ## is the PM2.5 **mass concentration in units of micrograms per cubic meter**. To reflect this in design and implementation, we need to correct the text in the hover in:

- [ ] the wireframes https://www.figma.com/proto/IZw8VvX1CsRc764CTvg4aRSU/WOAQ-Wireframes-Parent-File?node-id=753%3A813&scaling=contain&page-id=753%3A0
- [x] the app.

Also, 
- [x] Indicate units of measure at bottom of map or somewhere.

(The "moves with cursor" text in the first image corresponds to a different issue.)

![development app](https://user-images.githubusercontent.com/11000354/166612913-a1d23032-0143-4b07-a840-09bf53f19d68.png)
![wireframes](https://user-images.githubusercontent.com/11000354/166612917-50a4649f-d8f3-4901-b245-8bdda0cba8a2.png)

theecrit:
Add criteria:

- [ ] Indicate units of measure at bottom of map or somewhere.

gsehrlich:
Edited OP to include new criterion.

Issue #442, opened 2022-04-01
When hovering over a dot near the map's border, I'd like to the pop-up to appear fully within the map frame


theecrit:
**Is your feature request related to a problem? Please describe.**
Currently when I hover over the map near the edge, the data pop-up gets cut off by the frame and I can't view the whole thing.

**Describe the solution you'd like**
I'd like to be able to view the entire pop-up. One approach would be to detect which edge the pop-up is near and pop it up to either the right or left, depending on position.

**Describe alternatives you've considered**
We could try to display the pop-up over the extended edge of the map but that may be tough to do technically? 

**Screenshot
![Screen Shot 2022-04-01 at 3 05 15 PM](https://user-images.githubusercontent.com/32755410/161352556-dcf93b53-f310-4340-9c7f-0c29969415ac.png)
**

TangoYankee:
To achieve the left/right detection, attach a [element refs](https://learnreact.design/posts/react-useref-by-example) to the mapbox and the hover panel. Then, get their dimensions using `getBoundingClientRect`. 

To make the hoverpanel appear over the whole page: first, try giving it a z-index. If that doesn't work, Move it from being nested in the box and instead place it higher it in the DOM.

gsehrlich:
For a different issue ( the last piece of #331), I switched the hover implementation to use a react-map-gl element called Popup, which has the desired behavior by default. I'll reference this issue in the PR.

theecrit:
@gsehrlich  Amazing, thanks!

Issue #331, opened 2021-05-20
As a neighborhood citizen scientist or resident, I want to understand what a dot on the map represents


theecrit:
## Description
Each dot on the map represents a PM2.5 reading recorded at a specific time and place. Each dot will be color coded according to the EPA AQI (per #237). But as a citizen scientist, I may want more details, including:

- What the specific PM2.5 reading for that location is.
- What time of day the reading was recorded.
- Any other readings for that location (since the map defaults to displaying only the max reading for a given geopoint).
- Where the reading falls in terms of health risk.

## Acceptance criteria

The basic approach is to display a popup containing metadata when the viewer hovers over or clicks on (TBD) a specific geolocation marker. This popup should include:

- [x] All PM2.5 readings taken at the selected lon/lat point for the displayed collection session.
- [x] The time of day for each reading.
- [x] A color indicator for at least the max reading.
- [x] Add link to EPA-hosted information about AQI. (Jess to provide)

**Wireframes**
- See the [happy path prototype](https://www.figma.com/proto/IZw8VvX1CsRc764CTvg4aRSU/WOAQ-Wireframes-Parent-File?node-id=753%3A813&scaling=contain&page-id=753%3A0) for functionality.
- [Draft tooltip wires](https://www.figma.com/file/I2sQo358tSVKfkHiTpHoon/Map-Metadata?node-id=0%3A1) (option 3b)

## ~Design direction 2 (needs design exploration)~
~In this approach, the geolocation metadata would appear in a separate area of the interface (see example below). This might mean replacing the session metadata sidebar, or adding a new section below/above the map itself.~

~**Designs**~
~Example from Willet et al (note this issue doesn't cover any time-based representations):~

<img width="200" alt="Screen Shot 2021-06-30 at 10 35 24 AM" src="https://user-images.githubusercontent.com/32755410/124010756-15428400-d994-11eb-9bd8-4477911d78a3.png">

### Technical considerations
- [x] Assumes #237 is complete.
- [x] The primary sensor reading displayed should be pulled from most recent timestamp at that location.
- [x] “Other readings” should only appear if there are other readings.
- [x] “Other readings” should display in reverse chronological order.
- [ ] Popup and its contents must be discoverable by screen readers.
- [ ] Screen readers must be able to read back the content in proper order without confusing the user (for example, any color indicators need to be properly named and defined).
- [ ] For mobile users, tooltip should display on tap (w/ tap to remove).
[...]

PR #472, opened 2022-05-27
331 pin on click

gsehrlich:
## Checklist
- [x] Add description
  - lint `npm run lint-fix`
  - exit container as above
- [x] Request code review
- [x] Merge own code

## Description
Issues: #331, #442, and #464

- [x] Manually handle mouse interaction with map so that the popup follows the cursor while hovering over data and can be pinned by clicking on a data point
- [x] Add clickable question mark link to popup
- [x] Use react-map-gl's Popup element to fix the popup to a specific latitude and longitude; popup now automatically appears in optimal orientation to stay within the map frame
- [x] Give the popup a custom close button that plays nice with the new event handling
- [x] Splice in correct EPA link
- [x] When the user hovers over the question mark, show a modal at the top of map to clarify units and EPA link

Reviewer: It'll be easier to understand my edits by looking at the individual commits. At least for the last batch of commits (today's), I separated them into hunks representing atomic changes.

### DIFF
--- a/web/src/components/map/box.css
+++ b/web/src/components/map/box.css
@@ -9,3 +9,19 @@
     top: 10px;
     right: 10px;
 }
+
+.unit-info {
+    position: absolute;
+    top: 25px;
+    left: 0px;
+    right: 0px;
+    margin-left: auto;
+    margin-right: auto;
+    width: 290px;
+    background-color: rgba(0,0,0, 0.8);
+    color: #fff;
+    font-size: 11px;
+    text-align: center;
+    border-radius: 10px;
+    padding: 5px;
+}
--- a/web/src/components/map/box.css
+++ b/web/src/components/map/box.css
@@ -9,16 +9,3 @@
     top: 10px;
     right: 10px;
 }
-
- .hovertip {
-    position: absolute;
-    margin: 8px;
-    padding: 4px;
-    border-radius: 10px;
-    background: rgba(0,0,0, 0.8);
-    color: #fff;
-    max-width: 300px;
-    font-size: 10px;
-    z-index: 9;
-    pointer-events: none;
-}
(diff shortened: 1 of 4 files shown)


theecrit:
@gsehrlich Should we start with this link for the question mark icon destination? https://www.airnow.gov/aqi/aqi-basics/

It's imperfect, because it implies we're measuring more than just PM2.5, but I think it's a decent MVP start.

I also wonder if we should change the icon from a question mark to an off-site link icon. Normally I would discourage opening links in a new window, but given the nature of the interaction (that users will likely want to remain in the map experience), I wonder if we should make an exception in this instance. E.g. "external square alternate" from [Semantic UI](https://semantic-ui.com/elements/icon.html).

gsehrlich:
I think it's a decent start as well, certainly better than the EPA
homepage, which is what I have in there right now. My main concern is
actually that users will think the numbers on the map (which are in hidden
units of micrograms per cubic meter) mean the same thing as the numbers in
the chart (which have been converted to AQI). Especially since I haven't
added text anywhere yet to clarify that.

theecrit:
> My main concern is
actually that users will think the numbers on the map (which are in hidden
units of micrograms per cubic meter) mean the same thing as the numbers in
the chart (which have been converted to AQI).

Yeah, I hear that. Not sure we should worry about it for MVP but maybe prioritize it as a fast-follow for improvements?

theecrit:
Actually, I'm just realizing, we have an open criteria item in the comments on this in-progress issue: https://github.com/openoakland/woeip/issues/464#issuecomment-1123138453

gsehrlich:
Good point. Maybe I'll just implement that when I put the new link in and
call it done.

theecrit (approved):
LGTM!

